PR #30, opened 2021-05-05
Ryan add to analysis.py

rcanfield2:
Add functions to find the trough and peak time of a spike waveform


leoscholl (approved):
couple things:
- the function names don't really tell what they do, could you make more descriptive? at least have them start with a verb
- you should either add scipy to setup.py so that it gets installed by default, or have it surrounded by a try: catch: in case someone doesn't have it installed. in general we should also document this fact in the github sop or somewhere in the readthedocs

otherwise looks good

PR #31, opened 2021-05-05
example documentation

leoscholl:
added an example of how to preprocess a bmi3d task

### DIFF
--- a/aopy/data.py
+++ b/aopy/data.py
@@ -8,7 +8,7 @@ import csv
 import pandas as pd
 import os
 
-def get_filenames(base_dir, te):
+def get_filenames_in_dir(base_dir, te):
     '''
     Silly function to get the filenames for systems in a given task entry
 
@@ -19,7 +19,6 @@ def get_filenames(base_dir, te):
     Returns:
         dict: dictionary of files indexed by system
     '''
-    print("Please don't use this function. Make one that gets filenames from the database instead!")
     contents = os.listdir(base_dir)
     relevant_contents = [file_or_dir for file_or_dir in contents if str(te) in file_or_dir]
     files = {}
@@ -31,6 +30,18 @@ def get_filenames(base_dir, te):
         elif os.path.isdir(os.path.join(base_dir, file_or_dir)):
             files['ecube'] = file_or_dir
     return files
+
+def get_exp_filename(te):
+    '''
+    Returns the experiment filename for a given task entry
+
+    Args:
+        te (int): block number for the task entry
+    
+    Returns:
+        str: filename
+    '''
+    return "preprocessed_te" + str(te) + ".hdf"
 
 def load_optitrack_metadata(data_dir, filename, metadata_row=0):
     '''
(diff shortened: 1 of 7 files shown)


leoscholl:
sorry si jia i just pushed some new code by accident, please ignore the plotting code, i will have someone else review it

aopy/data.py
```diff
@@ -494,4 +572,64 @@ def load_bmi3d_root_metadata(data_dir, filename):
...
\ No newline at end of file
+        return dict(f['/'].attrs.items())
+
+_cached_signal_path = {}
+
+def lookup_acq2elec(data_dir, signal_path_file, acq, acq_name='acq', elec_name='electrode', zero_index=True):
+    '''
+    Looks up the electrode number for a given acquisition channel using an excel map file (from Dr. Map)
+
+    Inputs:
```
rcanfield2, line 583: 'Inputs' should be 'Args'

aopy/data.py
```diff
@@ -494,4 +572,64 @@ def load_bmi3d_root_metadata(data_dir, filename):
...
+        data_dir (str): where the signal path file is located
+        signal_path_file (str): signal path definition file
+        acq (int): which channel to look up
+        acq_name (str, optional): the name of the acquisition column
+        elec_name (str, optional): the name of the electrode column
+        zero_index (bool, optional): use 0-indexing for acq and elec (default True)
+
+    Output:
```
rcanfield2, line 591: 'Output' should be 'Returns'

aopy/data.py
```diff
@@ -494,4 +572,64 @@ def load_bmi3d_root_metadata(data_dir, filename):
...
+        ch2elec [nchannels]: lookup table mapping channels to electrodes or -1
+        pos (nelecs, nelecs): tuple of x and y positions for each electrode
+    '''
+    fullfile = os.path.join(data_dir, signal_path_file)
+    if fullfile in _cached_signal_path:
```
rcanfield2, line 596: Is '_cached_signal_path' a global variable? If should probably indicate this in the docstring somewhere and how to deal with it.

leoscholl: yeah, that way it doesn't have to load the excel file every time you call the function. i can mention that in the docstring, but i don't think the variable is something the user has to know about

rcanfield2: Sounds good. Just wanted to check. If the user doesn't need to know about it then it probably doesn't needed to added to the docstring

aopy/data.py
```diff
@@ -494,4 +572,64 @@ def load_bmi3d_root_metadata(data_dir, filename):
...
+        signal_path = _cached_signal_path[fullfile]
+    else:
+        signal_path = pd.read_excel(fullfile)
+        _cached_signal_path[fullfile] = signal_path
```
rcanfield2, line 600: When I tried to run the test code I was missing a package called 'xlrd'. From the error message it looks like it is required to use pd.read_excel so that package may need to be added to setup.py

aopy/data.py
```diff
@@ -494,4 +572,64 @@ def load_bmi3d_root_metadata(data_dir, filename):
...
+    Inputs:
+        data_dir (str): where the signal path file is located
+        signal_path_file (str): signal path definition file
+        acq (int): which channel to look up
+        acq_name (str, optional): the name of the acquisition column
+        elec_name (str, optional): the name of the electrode column
+        zero_index (bool, optional): use 0-indexing for acq and elec (default True)
+
+    Output:
+        ch2elec [nchannels]: lookup table mapping channels to electrodes or -1
```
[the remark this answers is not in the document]
rcanfield2: Also, this should probably indicate what it means when the function returns -1

aopy/data.py
rcanfield2: Also, if _cached_signal_path is a list of dataframes it might need a different name.

aopy/visualization.py
```diff
@@ -1,2 +1,165 @@
...
+        cmap (str, optional): matplotlib colormap to use in image
+    '''
+    # Calculate the proper extents
+    extent = [np.min(x), np.max(x), np.min(y), np.max(y)]
+    x_spacing = (extent[1]-extent[0])/(data_map.shape[0]-1)
+    y_spacing = (extent[3]-extent[2])/(data_map.shape[1]-1)
+    extent = np.add(extent, [-x_spacing/2, x_spacing/2, -y_spacing/2, y_spacing/2])
+
+    # Set the 'bad' color to something different
+    cmap = matplotlib.cm.get_cmap(cmap).copy()
```
pavi-rajes: Remove copy() 

This line below will work as well. 
cmap = matplotlib.cm.get_cmap(cmap) 

Copy() raises exceptions as we discussed. 
 line 157, in plot_spatial_map
    cmap = matplotlib.cm.get_cmap(cmap).copy()
AttributeError: 'LinearSegmentedColormap' object has no attribute 'copy'

aopy/visualization.py
```diff
@@ -1,2 +1,165 @@
...
+
+    data_map = np.reshape(X, grid_size)
+    return data_map
+
+def plot_spatial_map(data_map, x, y, ax=None, cmap='bwr'):
+    '''
+    Wrapper around plt.imshow for spatial data
+
+    Args:
+        map (2,n array): map of x,y data
```
pavi-rajes: map -> data_map just for consistency

tests/test_visualization.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+from aopy.visualization import *
+import numpy as np
+
+write_dir = 'tests/tmp'
```
pavi-rajes, line 5: not sure which is better - absolute path or relative path. 

I would add something like 
ROOT_DIR = os.path.abspath(os.curdir)
write_dir = os.path.join(ROOT_DIR, "tests/tmp")

aopy/visualization.py
```diff
@@ -1,2 +1,165 @@
...
+
+import matplotlib
+import matplotlib.pyplot as plt
+from scipy.interpolate import griddata
+from scipy.interpolate.interpnd import _ndim_coords_from_arrays
+from scipy.spatial import cKDTree
+import numpy as np
+import os
+
+def savefig(base_dir, filename, **kwargs):
```
pavi-rajes, line 12: We can add more functionalities to save_fig such as ability to save as a pdf. 
 
After some thoughts, I think it is useful to keep this function for saving. And add another function for figure settings. What do you think?

pavi-rajes (commented):
reviewed code for visualization and test_vizualization. Looks good. Here are my minor changes.

PR #36, opened 2021-05-13
Fix load bmi3d hdf

### LINKED ISSUE #33: data.load_bmi3d_hdf_table needs optional arg
Function: data.load_bmi3d_hdf_table

When the function isn't passed a value for the optional arg 'table_name' there is an error. (TypeError: getattr(): attribute name must be string).

leoscholl:
to fix #33


rcanfield2 (commented):
Looks good.

PR #37, opened 2021-05-16
Update preproc.py: Add condition handling for trial_separate and trial_align_events

rcanfield2:


tests/test_preproc.py
```diff
@@ -300,7 +302,8 @@ def test_trial_align_events(self):
             [0. , 0.1],
             [0. , 0. ],
```
leoscholl, line 303: i don't understand why there are some [0., 0.] rows in here, should they all be [0., 0.1]?

aopy/preproc.py
```diff
@@ -366,14 +367,24 @@ def trial_separate(events, times, evt_start, n_events=8):
...
+        if len(events[evt_start_idx[iE]: evt_start_idx[iE]+n_events]) == n_events:
+            trial_events[iE,:] = events[evt_start_idx[iE]: evt_start_idx[iE]+n_events]
+            trial_times[iE,:] = times[evt_start_idx[iE]: evt_start_idx[iE]+n_events]
+        else:
+            short_event_len = len(events[evt_start_idx[iE]: evt_start_idx[iE]+n_events])
+            trial_events[iE,:short_event_len] = events[evt_start_idx[iE]: evt_start_idx[iE]+short_event_len]
+            trial_events[iE, short_event_len:] = 0
+            trial_times[iE,:short_event_len] = times[evt_start_idx[iE]: evt_start_idx[iE]+short_event_len]
+            trial_times[iE,short_event_len:] = 0
+
```
leoscholl, line 380: i think this works for integer events but it won't work for string events. i'll make a commit with some suggested changes.

tests/test_preproc.py
rcanfield2: The [0,0] rows are where there are consecutive event codes to align to. Should probably be fixed though.

aopy/preproc.py
```diff
@@ -341,7 +341,8 @@ def calc_reward_rate(event_log, event_name='REWARD'):
 
 def trial_separate(events, times, evt_start, n_events=8):
     '''
-    Compute the 2D matrices contaning events per trial and timestamps per trial
+    Compute the 2D matrices contaning events per trial and timestamps per trial. 
+    If there are not enough events to fill n_events, the remaining indices will be missing values.
```
rcanfield2, line 345: Remaining indices will be missing values if the events are strings, -1 if the events are a signed int, and the maximum possible value for an unsigned int. Alternatively, within the function we should cast any unsigned int to a signed int since I can't think of a benefit to inputting an unsigned int.

PR #39, opened 2021-05-20
Add nevent_offset input to preproc.trial_separate

rcanfield2:


aopy/preproc.py
```diff
@@ -349,7 +349,8 @@ def trial_separate(events, times, evt_start, n_events=8):
         times (nt): times vector
         evt_start (int or str): event marking the start of a trial
         n_events (int): number of events in a trial
-    
+        nevent_offset (int): number of events to offset event alignment by
```
leoscholl, line 352: Could improve the documentation a little. maybe:
```suggestion
        nevent_offset (int): number of events before the start event to offset event alignment by. For example,
        if you wanted to align to "targ" in ["trial", "targ", "reward", "trial", "targ", "error"] but include the preceding "trial"
        event, then you could use nevent_offset=1
```

PR #41, opened 2021-05-21
add animate and frame event helper

leoscholl:
hi pavi, have a look at the sample_events function to see if it is useful for plotting events as rasters


leoscholl:
thanks for the suggestions, pavi!

PR #51, opened 2021-06-01
modify detect_edges

leoscholl:
not sure if check_alternating should be on by default. also whether that's a good name for it


m-nolan (approved):
Looks good, thank you for showing me the `|=` operator!

PR #55, opened 2021-06-03
Filter functions and plots 

pavi-rajes:
Hi, 

Please review the filter functions and plots from the files below:

precondition.py : 
- check code for filter functions 
- ideas for the plot we need to generate to test the quality of the filter 

test_precondition.py is not written like a unit test now. It is more like a unch of lines to generate plots now. Since the filter functions right now use functions from well-established libraries, I don't think we need unittests for the wrappers. But estimating multitaper-lfp_bandpower() will need a unittest case. Based on what you think, we can discuss the test functions we need.


tests/test_precondition.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # def test_butterworth(self):
+    #     # Sample rate and desired cutoff frequencies (in Hz).
+    #     # fs = 25000.0
+    #     lowcut = 500.0
+    #     highcut = 1200.0
+    #     tic = time.perf_counter()
+    #     self.x_filter = precondition.bandpass_butterworth_filter_data(_x, lowcut, highcut, fs)
+    #     toc = time.perf_counter()
+    #     print(f" Butterworth filter executed in {toc - tic:0.4f} seconds")
```
leoscholl, line 91:
```suggestion
    def setUp(self):
        self.T = 0.05
        self.fs = 25000
        self.freq = [600, 312, 2000]
        self.a = 0.02

        # testing generate test_signal
        _x, _t, _f0 = precondition.generate_test_signal(self.T, self.fs, self.freq, self.a)
        self.x = _x
        self.t = _t
        self.f0 = _f0

    def test_butterworth(self):
        # Sample rate and desired cutoff frequencies (in Hz).
        # fs = 25000.0
        lowcut = 500.0
        highcut = 1200.0
        tic = time.perf_counter()
        x_filter = precondition.bandpass_butterworth_filter_data(self.x, lowcut, highcut, self.fs)
        toc = time.perf_counter()
        fname = 'test_signal_filtered_Signal.png'
        print(f" Butterworth filter executed in {toc - tic:0.4f} seconds")
        precondition.plot_filtered_signal(self.t, self.x, x_filter, lowcut, highcut)
        plt.show()
        savefig(write_dir, fname)
```

leoscholl: this is how to set up the tests properly

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+        jackknife (bool): Use the jackknife method to make an estimate of the PSD variance at each point.
+        sides (str): This determines which sides of the spectrum to return.
+
+    Returns:
+        f (ndarray) : Frequency points vector
+        psd_est (ndarray): estimated power spectral density (PSD)
+        nu (ndarray): if jackknife = True; estimated variance of the log-psd. If Jackknife = False; degrees of freedom in a chi square model of how the estimated psd is distributed wrt true log - PSD
+    '''
+    f, psd_mt, nu = tsa.multi_taper_psd(data, fs, NW, BW,  adaptive, jackknife, sides )
+    return f, psd_mt, nu
```
leoscholl, line 102: either this function should be called get_psd_multitaper() or it should have arguments for low and high frequency cutoffs, then select the appropriate frequencies from the psd estimate, and transform back into time domain. i think both would be useful.

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+                lfp_power[idx * n_channels: (idx + 1) * n_channels, 0] = np.mean(np.log10(psd_est[fft_inds[idx]] + small_epsilon), axis=0)
+        else:
+            if no_log:
+                lfp_power[idx * n_channels: (idx + 1) * n_channels, 0] = np.mean(psd_est[:,fft_inds[idx]], axis=1)
+            else:
+                lfp_power[idx * n_channels: (idx + 1) * n_channels, 0] = np.mean(np.log10(psd_est[:,fft_inds[idx]] + small_epsilon), axis=1)
+
+    return lfp_power
+
+def get_psd(data, fs,l):
```
leoscholl, line 138: should probably be get_psd_welch() so we can later write a wrapper that has 'method' as a parameter

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+    '''
+    f, psd = signal.welch(data, fs, average='mean',nperseg=2*l)
+    return f,psd
+
+'''
+Plots to test filter performance
+'''
+
+
+def plot_filtered_signal(t,x, x_filter,low, high ):
```
leoscholl, line 159: these can go in visualization.py!
also i've been leaving out the plt.figure() part, so it will be possible to make subplots using these functions. you can follow the other examples in visualization.py

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+'''
+def generate_test_signal(T, fs, freq, a=0.02):
+    '''
+
+    Args:
+        T (float): Time period in seconds
+        fs (int): Sampling frequency in Hz
+        freq (1D array): Frequencies to be mixed in the test signal. main frequency in the first element
+        plot_true (bool): boolean value if true, plots the test signal in the provided axes
+        ax (plot axis handle): axis to plot the noisy test signal
```
rcanfield2, line 25: ax and plot_true don't need to be in the docstring

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+    # a = 0.02
+    f0 = freq[0]
+    # noise_power = 0.001 * fs / 2
+
+    x = a * 2 * np.cos(2 * np.pi * f0 * t)  # adding primary frequency
+    x += a * 0.5 * np.cos(2 * np.pi * freq[1] * t)
+    x += a * 1.5 * np.cos(2 * np.pi * freq[2] * t)
+    x += a * 20 * np.sin(2 * np.pi * 1.2 * np.sqrt(t))  # noise
+    # x += np.random.normal(scale=np.sqrt(noise_power), size=t.shape)  # noise
+
```
rcanfield2, line 41: Can this function only add 4 frequencies? It might be better to just make this a for loop for each input frequency with RNG terms as the amplitude.

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+    return x, t, f0
+
+def bandpass_butterworth_params(cutoff_low, cutoff_high, fs, order = 4, filter_type = 'bandpass'):
+    '''
+    Args:
+        cutoff_low (int): lower cut-off frequency (in Hz)
+        cutoff_high (int): higher cutoff frequency (in Hz)
+        fs (int): sampling rate (in Hz)
+        order (int): Order of the butter worth filter
+        filter_type (str) : Type of filter. Accepts one of the four values - {‘lowpass’, ‘highpass’, ‘bandpass’, ‘bandstop’}
```
rcanfield2, line 51: If the function is only designed to create bandpass params, filter_type doesn't need to be in the inputs.

aopy/precondition.py
```diff
@@ -1,2 +1,225 @@
...
+        nu (ndarray): if jackknife = True; estimated variance of the log-psd. If Jackknife = False; degrees of freedom in a chi square model of how the estimated psd is distributed wrt true log - PSD
+    '''
+    f, psd_mt, nu = tsa.multi_taper_psd(data, fs, NW, BW,  adaptive, jackknife, sides )
+    return f, psd_mt, nu
+
+def multitaper_lfp_bandpower(f,psd_est, bands, n_channels, no_log):
+    '''
+
+    Args:
+        psd_est (ndarray): power spectral density - output of bandpass_multitaper_filter_data
```
rcanfield2, line 108: missing f in the args section of the docstring

rcanfield2:
Looks pretty good overall. Just double check the docstrings are accurate for each function and add a 1-2 sentence description of what each function does before the 'Args' section of the docstring. Also, be sure to delete any code that is commented out because I assume it isn't needed. I'll think a little more about ideas for verifying/characterizing the filter.

aopy/precondition.py
pavi-rajes: renaming this function to get_psd_multitaper() to just get the power spectral density after tapers

aopy/precondition.py
pavi-rajes: removed bandpass from function names

aopy/analysis.py
```diff
@@ -3,11 +3,16 @@
 """
 
 import numpy as np
+from matplotlib import pyplot as plt
+import seaborn as sns
```
leoscholl, line 7: should seaborn be a dependency here?

aopy/analysis.py
```diff
@@ -41,26 +46,111 @@ def factor_analysis_dimensionality_score(data_in, dimensions, nfold, maxiter=100
...
+
+    for i in range(1, 9):
+        means_d.append(np.mean(data[target == i], axis=(0, 2)))
+        stds_d.append(np.std(data[target == i], axis=(0, 2)))
+
+    return means_d, stds_d
+
+
+def plot_mean_fr_per_target_direction(means_d, neuron_id, ax, color, this_alpha, this_label):
+    sns.set_context('talk')
```
leoscholl, line 101: can leave this out

aopy/analysis.py
```diff
@@ -41,26 +46,111 @@ def factor_analysis_dimensionality_score(data_in, dimensions, nfold, maxiter=100
...
+    stds_d = []
+
+    for i in range(1, 9):
+        means_d.append(np.mean(data[target == i], axis=(0, 2)))
+        stds_d.append(np.std(data[target == i], axis=(0, 2)))
+
+    return means_d, stds_d
+
+
+def plot_mean_fr_per_target_direction(means_d, neuron_id, ax, color, this_alpha, this_label):
```
leoscholl, line 100: plot_mean_fr should go in visualization.py

aopy/analysis.py
```diff
@@ -41,26 +46,111 @@ def factor_analysis_dimensionality_score(data_in, dimensions, nfold, maxiter=100
...
+    sns.set_context('talk')
+    ax.plot(np.array(means_d)[:, neuron_id], c=color, alpha=this_alpha, label=this_label)
+
+    ax.legend()
+    ax.set_xlabel("Target", fontsize=16)
+    ax.set_ylabel("Spike Rate (Hz)", fontsize=16)
+    plt.tight_layout()
+
+
+def run_curvefitting(means, make_plot=True, fig_title='Tuning Curve', n_rows=5, n_cols=11):
```
leoscholl, line 110: not sure if n_rows=5 and n_cols=11 makes sense as a default, since means could be any shape. this seems more specific to the analysis you were doing at the time than would make sense for aopy

maybe remove the plotting feature and just have this function do the curve fitting?

aopy/precondition.py
```diff
@@ -1,2 +1,161 @@
...
+import matplotlib.pyplot as plt
+#
+import nitime.algorithms as tsa
+import nitime.utils as utils
+from nitime.viz import winspect
+from nitime.viz import plot_spectral_estimate
+'''
+Filter functions
+'''
+def generate_test_signal(T, fs, freq, a):
```
leoscholl, line 17: should move to utils.py

aopy/precondition.py
leoscholl: i think your solution doesn't really make sense because you're always including cutoff_low and cutoff_high in the filter.

i feel like this is an instance where having a documentation page for filtering that explains how to use scipy.signal.butter would be better than writing this wrapper function

aopy/precondition.py
```diff
@@ -1,2 +1,161 @@
...
+        filter_type (str) : Type of filter. Accepts one of the four values - {‘lowpass’, ‘highpass’, ‘bandpass’, ‘bandstop’}
+
+    Returns:
+        tuple (b,a): bandpass filter parameters
+
+    '''
+    b,a = butter( order, [cutoff_low, cutoff_high], btype=filter_type, fs =fs)
+    return b,a
+
+def butterworth_filter_data(data, cutoff_low, cutoff_high, fs, order = 4,filter_type = 'bandpass' ):
```
leoscholl, line 62: also would remove filter_type from here and rename it to butterworth_bandpass_filter_data(), since it won't work if you only pass cutoff_low or vice versa.

aopy/precondition.py
leoscholl: f is still missing

aopy/precondition.py
```diff
@@ -1,2 +1,161 @@
...
+    return lfp_power
+
+def get_psd_welch(data, fs,l = None):
+    '''
+    Computes power spectral density using Welch's method
+
+    Args:
+        data (ndarray): time series data.
+        fs (float): sampling rate
+        l (int): no. of frequency points expected
```
leoscholl, line 147: could be more descriptive than 'l'
i suggest 'n_freq'

aopy/precondition.py
```diff
@@ -1,2 +1,161 @@
...
+        # average (str):  { ‘mean’, ‘median’ }, optional method to use when averaging periodograms. Defaults to ‘mean’.
```
leoscholl, line 148: delete this line

aopy/precondition.py
```diff
@@ -1,2 +1,161 @@
...
+            if no_log:
+                lfp_power[idx * n_channels: (idx + 1) * n_channels, 0] = np.mean(psd_est[:,fft_inds[idx]], axis=1)
+            else:
+                lfp_power[idx * n_channels: (idx + 1) * n_channels, 0] = np.mean(np.log10(psd_est[:,fft_inds[idx]] + small_epsilon), axis=1)
+
+    return lfp_power
+
+def get_psd_welch(data, fs,l = None):
+    '''
+    Computes power spectral density using Welch's method
```
leoscholl, line 142: you should explain that you're using the 'median' averaging method

aopy/visualization.py
```diff
@@ -1,17 +1,25 @@
...
+
 from scipy.interpolate import griddata
 from scipy.interpolate.interpnd import _ndim_coords_from_arrays
 from scipy.spatial import cKDTree
 import numpy as np
 import os
 import copy
+
+
+from aopy import precondition
```
leoscholl, line 17: this is incorrect, can't import aopy from within aopy
the correct line is 'from . import precondition'

aopy/visualization.py
```diff
@@ -47,6 +56,20 @@ def plot_timeseries(data, samplerate, ax=None):
...
+    ax.set_xlabel('Time (s)')
+    ax.set_ylabel('Voltage (uV)')
+
+
+def plot_freq_domain_power(data, samplerate, ax=None):
+    '''
+    Plots a power spectrum of each channel on the given axis
+    Args:
+        data (nt, nch): timeseries data, can also be a single channel vector
+    '''
```
leoscholl, line 72: did you add these lines by accident?

aopy/visualization.py
```diff
@@ -167,6 +194,104 @@ def plot_spatial_map(data_map, x, y, ax=None, cmap='bwr'):
     ax.set_xlabel('x position')
     ax.set_ylabel('y position')
 
+
+
+def plot_rastor(data, plot_cue, cue_bin, ax):
```
leoscholl, line 199: raster is spelled with an 'e'

aopy/visualization.py
```diff
@@ -167,6 +194,104 @@ def plot_spatial_map(data_map, x, y, ax=None, cmap='bwr'):
...
+           cue_bin : time bin at which an event occurs. For example: Go Cue or Leave center
+            ax: axis to plot rastor plot
+       Returns:
+           rastor plot in appropriate axis
+    '''
+    n_trial = np.shape(data)[0]
+    n_neurons = np.shape(data)[1]
+    n_bins = np.shape(data)[2]
+
+    color_palette = sns.set_palette("Accent", n_neurons)
```
leoscholl, line 215: i think these palette calls should be left up to the user, or make another function that makes these changes to your seaborn context

aopy/visualization.py
```diff
@@ -167,6 +194,104 @@ def plot_spatial_map(data_map, x, y, ax=None, cmap='bwr'):
...
+    for n in range(n_neurons):  # set this to 1 if we need rastor plot for only one neuron
+        for tr in range(n_trial):
+            for t in range(n_bins):
+                if data[n, tr, t] == 1:
+                    x1 = [tr, tr + 1]
+                    x2 = [t, t]
+                    ax.plot(x2, x1, color=color_palette(n))
```
leoscholl, line 222: this seems like it will be very slow. have you looked at https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.eventplot.html

aopy/visualization.py
```diff
@@ -167,6 +194,104 @@ def plot_spatial_map(data_map, x, y, ax=None, cmap='bwr'):
...
+    plt.show()
+
+
+def plot_db_spectral_estimate(freq, psd, psd_filter, labels):
+    psd = 10 * np.log10(psd)
+    psd_filter = 10 * np.log10(psd_filter)
+    plt.figure()
+    precondition.plot_spectral_estimate(freq, psd, (psd_filter,), elabels=(labels,))
+    plt.show()
+
```
leoscholl, line 294: i'm not seeing how these functions are broadly useful since they seem to output very specific plots. maybe more suitable for a jupyter notebook that you post in /examples/filtering.ipynb or something like that

aopy/analysis.py
```diff
@@ -41,26 +46,111 @@ def factor_analysis_dimensionality_score(data_in, dimensions, nfold, maxiter=100
...
+
+def get_mean_fr_per_direction(data, target):
+    means_d = []
+    stds_d = []
+
+    for i in range(1, 9):
+        means_d.append(np.mean(data[target == i], axis=(0, 2)))
+        stds_d.append(np.std(data[target == i], axis=(0, 2)))
+
+    return means_d, stds_d
```
leoscholl, line 97: these need docstrings and tests

tests/test_precondition.py
```diff
@@ -0,0 +1,88 @@
...
+class FilterTests(unittest.TestCase):
+
+    def __init_subclass__(self, **kwargs):
+        self.T = 0.05
+        self.fs = 25000
+        self.freq = [600, 312, 2000]
+        self.a = 0.02
+        self.f0 = self.freq[0]
+        # testing generate test_signal
+        _x, _t = precondition.generate_test_signal(self.T, self.fs, self.freq, self.a)
```
leoscholl, line 23: this can be deleted now that you have setUp()

tests/test_precondition.py
```diff
@@ -0,0 +1,88 @@
...
+        highcut = 1200.0
+        tic = time.perf_counter()
+        x_filter = precondition.butterworth_filter_data(self.x, lowcut, highcut, self.fs)
+        toc = time.perf_counter()
+        print(f" Butterworth filter executed in {toc - tic:0.4f} seconds")
+
+        fname = 'test_signal_filtered_Signal.png'
+        plot_filtered_signal(self.t, self.x, x_filter, lowcut, highcut)
+        plt.show()
+        savefig(write_dir, fname)
```
leoscholl, line 50: something i should also add to my tests in visualization.py -- a comment here saying what to expect from the figure so reviewers know what to check for

PR #58, opened 2021-06-08
Ryan add funcs to data.py and preproc.py

rcanfield2:
Add data.parse_str_list and preproc.locate_trials_with_event


aopy/preproc.py
```diff
@@ -619,6 +619,38 @@ def get_unique_trials(trial_idx, conditions, condition_name='target'):
                 corrected_trials = np.append(corrected_trials, trial)
     return corrected_trials
 
+def locate_trials_with_event(trial_events, event_codes, event_columnidx=None):
+    '''
+    Given an array of trial separated events, this function goes through and finds the event sequences corresponding to the trials
+    that include a given event. If an array of event codes are input, the function will find the trials corresponding to
+    each event code. 
+    Note: all events input through 'event_codes' must be located in the same column of 'trial_events'
```
rcanfield2, line 627: This line in the docstring is incorrect. I forgot to update it. If event_columnidx=None, the function will look at all columns

aopy/preproc.py
```diff
@@ -619,6 +619,38 @@ def get_unique_trials(trial_idx, conditions, condition_name='target'):
...
+    
+    Args:
+        trial_events (ntr, nevents): Array of trial separated event codes
+        event_codes (int, str, list, or 1D array): Event code(s) to find trials for. Can be a list of strings or ints
+        event_column (int): Column index to look for events in. Keep as 'None' if all columns should be analyzed.
+
```
pavi-rajes, line 633: might be worth mentioning event_column indexing starts from 0.  The first column should be represented as event_column_idx  = 0

aopy/preproc.py
```diff
@@ -619,6 +619,38 @@ def get_unique_trials(trial_idx, conditions, condition_name='target'):
...
+        
+    Returns:
+        (tuple):
+            (list of arrays): List where each index includes an array of trials containing the event_code corresponding to that index. 
+            (1D Array): Concatenated indices for which trials correspond to which event code.
+                        Can be used as indices to order 'trial_events' by the 'event_codes' input.
+    '''
+    split_events = []
+    if type(event_codes) == int or type(event_codes) == str:
+        split_events.append(np.unique(np.where(trial_events[:,event_columnidx] == event_codes)[0]))
```
pavi-rajes, line 642: Could use np.squeeze(np.where()) instead of np.unique(np.where()[0]).

split_events.append(np.squeeze(np.where(trial_events[:,event_columnidx]== event_codes)))

aopy/preproc.py
```diff
@@ -619,6 +619,38 @@ def get_unique_trials(trial_idx, conditions, condition_name='target'):
...
+        split_events=[np.array(split_events).flatten()]
```
pavi-rajes, line 643: I don't think this line is required. The above line is already appending a numpy array which is of 1 dimension. flatten() is not required.

aopy/preproc.py
pavi-rajes: I see you want to choose the unique elements. np.squueze(np.where()) might work for a single element. but unique() is required when event_codes is a list.  You can ignore my previous comment.

aopy/preproc.py
```diff
@@ -619,6 +619,38 @@ def get_unique_trials(trial_idx, conditions, condition_name='target'):
...
+        split_events_array = np.array(split_events).flatten()
+    else:
+        nevent_codes = len(event_codes)
+        split_events_array = np.array([]).astype(int)
+        for ievent in range(nevent_codes):
+            split_events.append(np.unique(np.where(trial_events[:,event_columnidx] == event_codes[ievent])[0]))
+            split_events_array = np.append(split_events_array, split_events[ievent])
+    
+    return split_events, split_events_array
+
```
pavi-rajes, line 653: what is split_events_array is not clear from the comment to me. I understand split_events is actually trial index with event.  split_events_array is a flattened version of split_events without which event_code it belongs to ? 
When would it be useful? Should the function return it (or flatten() could be applied outside the function)?

pavi-rajes (commented):
Code looks clean. Both tests passed when I ran locally. Suggesting minor changes only.

PR #64, opened 2021-06-16
Leo fixes

leoscholl:
hi ryan, this pull request addresses a couple issues you raised. could you check that i fixed everything?


rcanfield2 (commented):
Looks good. I just added seaborn to the setup.py file since it was missing.

PR #71, opened 2021-07-01
add posmap to docs

leoscholl:
testing whether we can add plots to visualization.rst

if it looks good we can merge to master and see what the docs look like


pavi-rajes (commented):
format of doc string for code block and image looks good.

PR #74, opened 2021-07-12
Add first function needed to compute fano factor

rcanfield2:


m-nolan (commented):
Looks like a standard Fano Factor calcuation. Approved

m-nolan (approved):

PR #82, opened 2021-07-12
Ryans lab meeting functions: fano factor values

rcanfield2:


aopy/analysis.py
```diff
@@ -292,3 +292,23 @@ def interpolate_extremum_poly2(extremum_idx, data, extrap_peaks=False):
 
     return extremum_time, extremum_value, f
 
+def get_fano_factor_values_per_condition(spiking_data):
```
rcanfield2, line 295: Make function more generic because this isn't actually calculating the fano factor

aopy/analysis.py
```diff
@@ -292,3 +292,23 @@ def interpolate_extremum_poly2(extremum_idx, data, extrap_peaks=False):
...
+    '''
+    This function calculates the two parameters used to calculate fano factor for input spiking data based on Churchland et al. 2010.
```
rcanfield2, line 297: Put a doi to Churchland et al. 2010 to ensure the right one is found.

aopy/analysis.py
```diff
@@ -292,3 +292,23 @@ def interpolate_extremum_poly2(extremum_idx, data, extrap_peaks=False):
...
+    These two parameters are calculated for each unit and are the mean spikes per trial and the spiking variance of each unit across trials.
+
```
rcanfield2, line 299: Show documentation about where this is used

aopy/analysis.py
```diff
@@ -292,3 +292,23 @@ def interpolate_extremum_poly2(extremum_idx, data, extrap_peaks=False):
...
+    Args:
+        spiking_data (ntime, nunits, ntr): Input spiking data
+        weight_regression (bool): A flag to weight the linear regression based on the number of trials.
```
rcanfield2, line 302: remove this variable input.

aopy/analysis.py
rcanfield2: Add assumptions made by function. If not here than up stream.

leoscholl (approved):

PR #89, opened 2021-07-13
maneeshika-branch: added kalman filter function to analysis.py

mmadduri:
added import in numpy and function, no testing added


zaicheek (changes requested):
please add appropriate tests

PR #94, opened 2021-07-14
Mnolan dev

m-nolan:
Add some pytorch dataset code to data.py. Includes documentation (should work).


docs/source/utils.rst
```diff
@@ -2,6 +2,8 @@ Utils:
 ===============
 
 Utility functions for aopy data management and analysis
+
+Utility Functions
```
leoscholl, line 6:
```suggestion
```

leoscholl (commented):
you should add 'torch' to the list of dependencies in `setup.py`
and also in the mock imports in `docs/conf.py`

aopy/data.py
```diff
@@ -744,3 +745,121 @@ def parse_str_list(strings, str_include=None, str_avoid=None):
...
+                f.write(np.byte(intdata[t,ch] >> 8))
+
+    return filename
+
+# - - -- --- ----- -------- ---PYTORCH--- -------- ----- --- -- - -
+# - - -- --- ----- -------- ---DATASETS-- -------- ----- --- -- - -
+
+class EcogTensorDataset(torch.utils.data.Dataset):
+    r"""Dataset wrapping tensors of ECoG data. Implements the pytorch Dataset parent class.
+
```
leoscholl, line 809: Seems like a better name for this might be TransformableDataset or something along those lines.. why is this specific to ECoG data?

leoscholl (approved):
looks good after you address my comments

PR #103, opened 2021-07-19
Add torch to conf.py

rcanfield2:


m-nolan (approved):
This was my mistake. The package isn't called `pytorch`. Thanks for catching that.

PR #112, opened 2021-07-30
Kalman Filter functions

pavi-rajes:
I have included a simple implementation of Kalman Filter in analysis.py. Could you review the function and associated docstrings, especially the math equations? I plan to include an example with plots in an rst file. after checking this implementation once with Amy.


leoscholl (commented):
i think you need to write some tests (you can use one as your example) before we can properly review.

zaicheek (approved):
matrix operations changed to np.matmul style (@)

PR #113, opened 2021-07-30
Leo plotting

leoscholl:
changed one function plot_columns_by_date() -> plot_sessions_by_date(), removing the need to have a dataframe as input

added two functions, plot_sessions_by_trial() and color_trajectories()


zaicheek (approved):
plot_sessions is more flexible that plot_columns. thank you for adding color_trajectories() specifically. everything looks good, i only cleaned up the merge conflicts.

PR #117, opened 2021-08-10
Extract clda table and save to processed files

sijia66:


leoscholl (commented):
only slight change i would make is to include _bmi3d_ in the name of the function that checks if the table exists. maybe bmi3d_hdf_contains()?

aopy/data.py
```diff
@@ -621,6 +621,22 @@ def load_bmi3d_root_metadata(data_dir, filename):
...
+def is_table_in_hdf(table_name:str, hdf_filename:str):
+    """
+    Checks if a table exists in an hdf file' first level directory(i.e. non-recursively)
+
+    Args:
+        table_name(str): table name to be checked
+        hdf_filename(str): full path to the hdf file
+    
+    Returns: 
+        Boolean
```
leoscholl, line 634:
```suggestion
        bool: whether the hdf file contains the given table or not
```

leoscholl (approved):

PR #120, opened 2021-08-25
update get_pca_dimensions

rcanfield2:
- Fix test for this function. 
- Removed the 'dimensions' output because I wasn't sure how it was different than the features of data you input. 
- Removed the for loop to decrease computation time.


zaicheek (commented):
this looks good. i think we can eliminate dimensions, it was only being used as a placeholder to implement through a for loop and you used some functions i just learned about to get around it! :)

zaicheek (approved):

PR #122, opened 2021-08-27
move torch to separate folder, not required

leoscholl:
is this ok? moving the torch stuff to a different file means it doesn't have to be required for install, i think. since pytorch is quite a large package (~1GB for me)

### DIFF
--- a/aopy/data.py
+++ b/aopy/data.py
@@ -1,7 +1,6 @@
 # data.py
 # code for accessing neural data collected by the aoLab
 import numpy as np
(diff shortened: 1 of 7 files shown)


rcanfield2 (commented):
I think this is okay as well but the documentation should also probably be moved and a note added saying that torch is required to be installed to run the functions in this module.

leoscholl:
done. can one of you approve the changes?

m-nolan (approved):

PR #124, opened 2021-09-04
Load ecube chunks

leoscholl:
i added a function to load an ecube file one chunk at a time, where the chunk size is a number of samples. i will use this function in bmi3d to simulate data being streamed by the ecube, which arrives in 728-sample packets. 

let me know what you think.

### DIFF
--- a/aopy/data.py
+++ b/aopy/data.py
@@ -254,11 +254,10 @@ def load_ecube_data(data_dir, data_source, channels=None):
 
     # Fetch all the data for all the channels
     n_samples = metadata['n_samples']
-    n_channels = metadata['n_channels']
-    timeseries_data = np.zeros((n_samples, n_channels), dtype=dtype)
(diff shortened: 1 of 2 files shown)


m-nolan (approved):
I like the move to the generator construction in _process_channels(). Very python-y! The transition to the new generator function looks complete and the direct access to the chunked outputs looks like it will work. I don't know what the syntax is for returning a generator like range() or enumerate() do, but the first assert in the test method won't pass unless it's working as intended (the next() function).

leoscholl:
> the first assert in the test method

i'm not sure which test method you're referring to, can you be more specific?

m-nolan:
The `next()` function is only going to work if you've returned a valid generator object, right? My confusion was about how the `_process_channels()` function actually returned data, but the test will only pass if `next` is called correctly, so that makes me think that it's supposed to be returning a generator and that's OK. It means "I'm confused about the innards but the tests show how it's supposed to be used."

PR #125, opened 2021-09-06
add bmi3d v6

leoscholl:


rcanfield2 (approved):
Looks good to me but its hard to be 100% certain without knowing much about what changed in sync_protocol_version 6

PR #130, opened 2021-09-10
ryan misc funcs

rcanfield2:


aopy/analysis.py
```diff
@@ -348,3 +349,26 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9):
...
+    from the center point are labeled as outliers. 
+    
+    Inputs:
+        data [n, nfeatures]: 
+        std_threshold [float]: Number of standard deviations away a data point is required to be to be classified as an outlier
+        
+    Outputs: 
+        good_data_idx [n]: Labels each data point if it is an outlier (True = good, False = outlier)
+        distances [n]: Distance of each data point from center
+    '''
```
leoscholl, line 365: can you include the examples from your tests in the docstrings?

aopy/analysis.py
```diff
@@ -348,3 +349,26 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9):
...
     return dimensions, explained_variance, num_dims
 
+def find_outliers(data, std_threshold):   
+    '''
+    Use kmeans clustering to find the center point of a dataset and distances from each data point
+    to the center point. Data points further than a specified number of standard deviations away
+    from the center point are labeled as outliers. 
+    
+    Inputs:
+        data [n, nfeatures]:
```
leoscholl, line 359: missing description

aopy/analysis.py
```diff
@@ -348,3 +349,26 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9):
 
     return dimensions, explained_variance, num_dims
 
+def find_outliers(data, std_threshold):
```
leoscholl, line 352: kind of a weird name for the function. since it doesn't only find the outliers, it also finds the clusters. maybe find_clusters? or find_clusters_and_outliers? or label_clusters?

aopy/data.py
```diff
@@ -743,3 +743,35 @@ def parse_str_list(strings, str_include=None, str_avoid=None):
...
             
     return parsed_str
+
+def load_matlab_cell_strings(data_dir, hdf_filename, object_name):
+    '''
+    This function extracts strings from an object within .mat file that was saved from 
+    matlab in version -7.3 (-v7.3). 
+
+    Args:
+        data_dir (str): where the signal path file is located
```
leoscholl, line 753:
```suggestion
        data_dir (str): where the matlab file is located
```

leoscholl (approved):

aopy/analysis.py
rcanfield2: It actually only uses the kmeans function to find the center point of the data if it were one cluster, therefore all data poitns are in the same cluster and the cluster info isn't useful. I'll add this info to the docstring and that this is particularly useful for high dimensional data.

PR #132, opened 2021-09-14
Leo add rms

leoscholl:


aopy/analysis.py
```diff
@@ -337,3 +337,19 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9):
...
 
+def calc_rms(signal, remove_offset=True):
+    '''
+    Root mean square voltage of a signal
+    
+    Args:
+        signal (nt, ...): voltage along time, other dimensions will be preserved
+
+    Returns:
+        float: rms of the signal along the first axis
```
rcanfield2, line 348: Wont this have same non-time dimensions as the input signal?

aopy/analysis.py
```diff
@@ -337,3 +337,19 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9):
 
     return list(explained_variance), num_dims
 
+def calc_rms(signal, remove_offset=True):
+    '''
+    Root mean square voltage of a signal
+    
+    Args:
+        signal (nt, ...): voltage along time, other dimensions will be preserved
```
rcanfield2, line 345: Add documentation for the remove_offset input variable.

aopy/analysis.py
```diff
@@ -337,3 +337,19 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9):
 
     return list(explained_variance), num_dims
 
+def calc_rms(signal, remove_offset=True):
+    '''
+    Root mean square voltage of a signal
```
rcanfield2, line 342: Doesn't necessarily need to be the voltage.

tests/test_analysis.py
```diff
@@ -106,7 +106,19 @@ def test_get_pca_dimensions(self):
...
+        # sanity check
+        signal = np.array([1])
+        rms = aopy.analysis.calc_rms(signal)
+        self.assertEqual(rms, 0)
+
+        # check dimensions are preserved
+        signal = np.array([[0.5, -0.5], [1., -1.]]).T
+        rms = aopy.analysis.calc_rms(signal)
+        self.assertCountEqual(rms, np.array([0.5, 1.]))
+
```
rcanfield2, line 121: I think it would be good to check that 'remove_offset=False' works in one of these two checks.

tests/test_analysis.py
```diff
@@ -106,7 +106,19 @@ def test_get_pca_dimensions(self):
...
+    def test_calc_rms(self):
+        # sanity check
+        signal = np.array([1])
+        rms = aopy.analysis.calc_rms(signal)
+        self.assertEqual(rms, 0)
+
+        # check dimensions are preserved
+        signal = np.array([[0.5, -0.5], [1., -1.]]).T
+        rms = aopy.analysis.calc_rms(signal)
+        self.assertCountEqual(rms, np.array([0.5, 1.]))
```
rcanfield2, line 120: If I understand correctly this test ensures the two sequences contain the same elements regardless of order. I think it would be better to use an order preserving check like np.testing.assert_allclose

rcanfield2 (approved):

PR #146, opened 2021-09-18
drmap funcs

rcanfield2:


aopy/data.py
```diff
@@ -702,6 +702,65 @@ def load_electrode_pos(data_dir, pos_file):
     y_pos = electrode_pos['topdown_y'].to_numpy()
     return x_pos, y_pos
 
+def map_acq2pos(config_dir, signalpath_filename, elec2pos_filename):
+    '''
+    Create index mapping from acquisition channel to electrode position. 
+    Input file types must be compatible with pd.read_excel. 
+    Electrode position column names must be 'topdown_x' and 'topdown_y'
```
leoscholl, line 709: probably should have these be arguments with defaults topdown_x and topdown_y

aopy/data.py
```diff
@@ -702,6 +702,65 @@ def load_electrode_pos(data_dir, pos_file):
...
+    Args:
+        configdir (str): Directory holding the configuration files
+        signalpath_filename (str): Filename of the signal path information (Mapping between electrode and acquisition ch)
+        posmap_filename (str): Filenmae of the electrode position information. (Mapping between electrode and position on array)
+    
+    Returns:
+        acq_ch_position (nelec, 2): X and Y coordinates of the electrode each acquisition channel gets data from.
+                                    X position is in the first column and Y position is in the second column
+        acq_chs (nelec): Acquisition channels that map to electrodes (240/264)
+        connected_elecs (nelec): Electrodes used (240/244)
```
leoscholl, line 720:
```suggestion
        acq_chs (nelec): Acquisition channels that map to electrodes (e.g. 240/256 for viventi ECoG array)
        connected_elecs (nelec): Electrodes used (e.g. 240/244 for viventi ECoG array)    
```

tests/test_data.py
```diff
@@ -269,7 +270,42 @@ def test_load_electrode_pos(self):
         x, y = load_electrode_pos(data_dir, testfile)
         self.assertEqual(len(x), 244)
         self.assertEqual(len(y), 244)
-    
+
+    def test_map_acq2pos(self):
+        test_signalpathfile = '210910_ecog_signal_path.xlsx'
+        test_mappingfile = '244ch_viventi_ecog_elec_to_pos.xls'
```
leoscholl, line 276: i would call it a "layout file" so as not to confuse with other maps
```suggestion
        test_layoutfile = '244ch_viventi_ecog_elec_to_pos.xls'
```

leoscholl (commented):
this is great, nicely done. a couple suggestions:

- it might be useful to have the functions work with a subset of the acquisition channels instead of the whole dataset. can you add an optional parameter which lets you select the acquisition channels being used, maybe this can default to all the channels
- i like how the `map_data2elec` reuses the `map_acq2pos` function, but it seems unnecessary to require a spatial layout (elec2pos) in all cases. i would add a third function called `map_acq2elec` which does the bulk of the work, then `map_acq2pos` just adds the spatial layout on top, and `map_data2elec` applies the map to data without needing to deal with spatial layouts.

rcanfield2:
Included a way to select acq channels as an input to the function but I could see it also being useful to be able to select based on position. I think adding that functionality can be another issue/pull request.

leoscholl:
could you also change the channel map inputs to be tables rather than filenames? 

you can add another function `load_signal_path_file()` or something with the line:

```
# Load signal path data as a pandas dataframe
    return pd.read_excel(os.path.join(config_dir, signal_path_filename))
```

that way we can use the mapping functions with dataframes saved from BMI3D metadata

rcanfield2:
replace map_data2elec with 2 functions, one thats a comprehensive wrapper and one that actually maps data 2 elec

rcanfield2:
Decided to not write a function to load the excel files because it is just one line and the user can just call pd.read_excel. I did include a reference to this in the docstrings of each function.

aopy/data.py
```diff
@@ -702,6 +703,112 @@ def load_electrode_pos(data_dir, pos_file):
...
+    acq_chs = signalpath_table['acq'][connected_elecs_mask].to_numpy(dtype=int)
+
+    if acq_ch_subset is not None:
+        acq_chs_mask = np.where(np.in1d(acq_chs, acq_ch_subset))[0]
+        acq_chs = acq_chs[acq_chs_mask]
+        connected_elecs = connected_elecs[acq_chs_mask]
+        if len(acq_chs) < len(acq_ch_subset):
+            missing_acq_chs = acq_ch_subset[np.in1d(acq_ch_subset,acq_chs, invert=True)]      
+            warning_str = "Requested acquisition channels " + str(missing_acq_chs) + " are not connected"
+            warnings.warn(warning_str)
```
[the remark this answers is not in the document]
rcanfield2: That was the impetus for including it. It doesn't break/stop the function, it just displays a warning that some requested channels (and which ones) aren't connected since the user can't know beforehand.

aopy/data.py
```diff
@@ -702,6 +703,112 @@ def load_electrode_pos(data_dir, pos_file):
...
+        eleclayout_table (pd dataframe): Electrode position information in a pandas dataframe. (Mapping between electrode and position on array)
+        acq_ch_subset (nacq): Subset of acquisition channels to call. If not called, all acquisition channels and connected electrodes will be return. If a requested acquisition channel isn't returned a warned will be displayed
+        xpos_name (str): Column name for the electrode 'x' position. Defaults to 'topdown_x' used with the viventi ECoG array
+        ypos_name (str): Column name for the electrode 'y' position. Defaults to 'topdown_y' used with the viventi ECoG array
+
+    Returns:
+        acq_ch_position (nelec, 2): X and Y coordinates of the electrode each acquisition channel gets data from.
+                                    X position is in the first column and Y position is in the second column
+        acq_chs (nelec): Acquisition channels that map to electrodes (e.g. 240/256 for viventi ECoG array)
+        connected_elecs (nelec): Electrodes used (e.g. 240/244 for viventi ECoG array)
```
leoscholl, line 751: how should we index acq channels and electrodes? does it make a difference if we use 1-indexing or 0-indexing? maybe a note about using the same indexing scheme as the signalpath table? or can we enforce it either way?

aopy/data.py
```diff
@@ -702,6 +703,112 @@ def load_electrode_pos(data_dir, pos_file):
...
+        acq_ch_subset (nacq): Subset of acquisition channels to call. If not called, all acquisition channels and connected electrodes will be return. If a requested acquisition channel isn't returned a warned will be displayed
+
+    Returns:
+        dataout (nt, nelec): Data from the connected electrodes
+        acq_chs (nelec): Acquisition channels that map to electrodes (e.g. 240/256 for viventi ECoG array)
+        connected_elecs (nelec): Electrodes used (e.g. 240/244 for viventi ECoG array) 
+    '''
+    
+    acq_chs, connected_elecs = map_acq2elec(signalpath_table, acq_ch_subset=acq_ch_subset)
+    dataout = datain[:,acq_chs-1]
```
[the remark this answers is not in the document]
rcanfield2: input it as a nt x nacqch shaped array

aopy/data.py
rcanfield2: I can add an optional argument to set to true if the acquisition channels are 0-indexed. It will default to False indicating 1-indexing.

leoscholl (approved):

PR #153, opened 2021-09-23
refactor modules

leoscholl:
add a message to the top of each module explaining what it's for, also move a couple functions around

### DIFF
--- a/aopy/data.py
+++ b/aopy/data.py
@@ -9,6 +9,7 @@ import pandas as pd
 import os
 import glob
 import warnings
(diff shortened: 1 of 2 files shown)


leoscholl:
i moved a bunch of things from preproc to utils, i think you were right about putting more math/generic functions there. one of the functions literally said "Utility for ..." in the first line of the docstring.

the only module that is still a bit broken is postproc. at least some of those functions definitely belong in preproc, but i'm not sure where to draw the line.

PR #157, opened 2021-09-25
initial eye tracking uploading

leoscholl:


aopy/preproc.py
```diff
@@ -1374,3 +1374,187 @@ def proc_lfp(data_dir, files, result_dir, result_filename, overwrite=False):
...
+        'left_eye':left_eye_caliberation,
+        'right_eye':right_eye_caliberation
+    }
+    
+    return caliberation_file
+    
+    
+
+    
+def calc_single_eye_caliberation(cursor_trajectories_aligned, eye_aligned):
```
leoscholl, line 1496: could you make this into a more general function and put it into analysis.py? basically it takes two 2D trajectories and computes the linear regression between their x and y coordinates. doesn't have to be eye and cursor data, necessarily.

since you're having issues with the save_hdf() and dictionaries of dictionaries, maybe it would be better to return the x and y coefficients together. something like 

```
coefficients = np.array([[slope_x, intercept_x], [slope_y, intercept_y]]),
correlation_coeff = np.array([r_value_x, r_value_y])
```

then you could save four variables into the hdf file: left_eye_coefficients, right_eye_coefficients, left_eye_correlation_coeff, right_eye_correlation_coeff

sijia66: it should save and added caliberate funcs, 
should work, 
need to test on a few more days of  data.

sijia66 (approved):

Issue #84, opened 2021-07-13
add function to bin discrete events


leoscholl:
take in spike data; return binned spike rates in time. This should also work if data has already been trial aligned.

args: 
    spike times (nt, nch, (ntrial))
    bin width (time)
    
returns:
    (nbin, nch, (ntrial)) - binned spikes

rcanfield2:
Write a basic function to bin spike data. 'waveform_data' is an (nt x 1) array of data. It would be useful to expand this function to accommodate higher dimensional arrays.

```
def bin_spikes(waveform_data, nbins, bin_width, total_time_s):
    binned_unit_spikes, _ = np.histogram(waveform_data, bins=nbins, range=(0, total_time_s))
    binned_unit_spikes = binned_unit_spikes/bin_width # convert from [spikes/bin] to [spikes/s]
    return binned_unit_spikes
```

PR #158, opened 2021-09-26
I made code for binning spikes

resatomo:
[#84](https://github.com/aolabNeuro/analyze/issues/84) 
As Ryan suggested, this code bins 1d spike data.

### DIFF
--- a/aopy/precondition.py
+++ b/aopy/precondition.py
@@ -162,5 +162,29 @@ def get_psd_welch(data, fs,n_freq = None):
         f, psd = signal.welch(data, fs, average = 'median')
     return f,psd
 
+def bin_spikes(data, fs, bin_width):
+    '''
+    bins spikes
 
+    Args:
+        data (ndarray): time series spike data (nt x 1).
+        fs (float): sampling rate of data
+        bin width (int): width of bin
+
+    Returns:
+        binned_spikes (ndarray): binned spikes 
+    '''
+
+    dT = 1/fs # time width of bin
+    nbins = math.ceil(len(data)/bin_width) # the number of bins
+    binned_spikes = np.zeros(nbins)
+
+    for idx in range(nbins):
+        if idx == nbins - 1:
+            binned_spikes[idx] = np.mean(data[idx*bin_width :])
+        else:
+            binned_spikes[idx] = np.mean(data[idx*bin_width : (idx + 1)*bin_width])
+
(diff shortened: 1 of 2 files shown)


leoscholl (commented):
this looks pretty good, i think we should try to make it compatible with multiple channels though. 

also this isn't quite what i had in mind, since you're binning the sampled thresholded data instead of spike times, but i think it's fine.

aopy/precondition.py
```diff
@@ -162,5 +162,29 @@ def get_psd_welch(data, fs,n_freq = None):
         f, psd = signal.welch(data, fs, average = 'median')
     return f,psd
 
+def bin_spikes(data, fs, bin_width):
+    '''
+    bins spikes
+
```
leoscholl, line 168: Could you add the example here to make it obvious what the inputs/outputs are

aopy/precondition.py
```diff
@@ -162,5 +162,29 @@ def get_psd_welch(data, fs,n_freq = None):
...
+        data (ndarray): time series spike data (nt x 1).
+        fs (float): sampling rate of data
+        bin width (int): width of bin
+
+    Returns:
+        binned_spikes (ndarray): binned spikes 
+    '''
+
+    dT = 1/fs # time width of bin
+    nbins = math.ceil(len(data)/bin_width) # the number of bins
```
leoscholl, line 179: if you use data.shape[0] instead of len(data) it will be more clear which length you are referring to

aopy/precondition.py
```diff
@@ -162,5 +162,29 @@ def get_psd_welch(data, fs,n_freq = None):
...
+    binned_spikes = np.zeros(nbins)
```
leoscholl, line 180: here you can adjust the shape to be `np.zeros((nbins, nch))` where nch is the size of the second dimension `data.shape[1]`

aopy/precondition.py
```diff
@@ -162,5 +162,29 @@ def get_psd_welch(data, fs,n_freq = None):
...
+
+    for idx in range(nbins):
+        if idx == nbins - 1:
+            binned_spikes[idx] = np.mean(data[idx*bin_width :])
```
leoscholl, line 184: then adjust these to apply the mean along axis 0 and put it into `binned_spikes[idx,:]`

leoscholl (commented):
hi tomo, just one more suggestion: can you change the bin_width argument to have units of seconds instead of samples? i think this will make it easier to use. i made a few minor changes to the test function to make it obvious what i mean

PR #167, opened 2021-10-18
detect spikes code and plot

rcanfield2:


tests/test_precondition.py
```diff
@@ -141,6 +141,38 @@ def test_multitaper(self):
...
+        for ii in range(large_data.shape[1]):
+            large_data[:,ii] = np.arange(large_data.shape[0])
+            
+        large_data[10:,0] = np.arange(0,large_data.shape[0]-10) 
+
+        _, wfs = precondition.detect_spikes(large_data, 300, wf_length=10000, threshold=threshold)
+        np.testing.assert_allclose(wfs[0], np.array(((3,4,5),(3,4,5))))
+        np.testing.assert_allclose(wfs[1], np.array((8,9,10)).reshape(1,-1))
+        np.testing.assert_allclose(wfs[2], np.array((13,14,15)).reshape(1,-1))
+        np.testing.assert_allclose(wfs[3], np.array((np.nan,np.nan,np.nan)).reshape(1,-1))
```
leoscholl, line 174: i would also add one more test with more substantial test data to make sure it's not super slow or anything. if you just make some gaussian noise it should give you some "spikes" back. a reasonably sized chunk of data for our use case is something like 256 channels by 250,000 samples (10 seconds)

Issue #133, opened 2021-09-15
plot_spatial_map error with single channel


leoscholl:
if given a single datapoint and single x and y pos, then we get the error: `ValueError: Axis limits cannot be NaN or Inf`

maybe it doesn't make sense to plot a single channel spatial map? but i think for compatibility it should work.

PR #168, opened 2021-10-18
I changed a part of function Plot_spatial_map so that it can plot sin…

resatomo:
[#133](https://github.com/aolabNeuro/analyze/issues/133) 
In the single-channel case, when computing `extent` which is the argument of `imshow`, its value was infinity. So I prepared exceptional `extent` when computing the single-channel mapping.


leoscholl (approved):
looks good, just a few minor suggestions and then you can merge

PR #171, opened 2021-10-20
Update detect spikes

rcanfield2:
Update detect spikes to accommodate negative spikes.


rcanfield2:
Oops yeah the automatic negative thresholding is wrong. I'll pull out the automatic thresholding as you suggested. However, I don't think taking the absolute value of the signal works. If the threshold is 0.5 and the signal jumps from 1 to -1, there wouldn't be an edge to detect since the signal is rectified. I think i would need to detect positive and negative spikes individually and then combine them.

leoscholl:
maybe that points at a bigger problem of using the detect_edges() function when you should really just find any values above or below the threshold?

rcanfield2:
TODO:
- Add mean-RMS as an option to threshold calculation
- Add refractory period to spike detection
- Add waveform length offset so the user can grab samples before the detected spike time.

rcanfield2:
The spike detection with refractory period still has a slight problem. Say the refractory period is 1 second. If a spike occurs at 1, 1.5, and 2.25 seconds, both of the final spikes will be tossed out as bad spikes when we may want to keep the last one. The possible solutions I though of are:

1. Keep the function as is. With a small enough refractory period this will be negligible.
2. Write a for loop to check each spike. This seems very slow. 
3. Write a for loop to only check bad spikes. This requires substantial logic to check.

We can discuss tomorrow but I wanted to note these thoughts before I forget.

rcanfield2:
@leoscholl

leoscholl (approved):

PR #179, opened 2021-10-29
Fix curve fitting format

pavi-rajes:
Addressing Leo's comments and adding references to relevant other functions in doctoring to curve fitting functions in analysis.py.


pavi-rajes:
not sure why I have Kaman filter commits also stacked on my commits about doctoring update. Will fix that and create a pull request for doctoring update only.  Ignoring this pull request for now.

PR #182, opened 2021-11-02
add tests for ecube data gen

leoscholl:
i added more comprehensive tests for the ecube data generation / loading since i wasn't getting the same thing back when i loaded a test signal.

but there doesn't seem to be anything wrong with it.. i'm stumped

### DIFF
--- a/aopy/utils.py
+++ b/aopy/utils.py
@@ -64,6 +64,12 @@ def save_test_signal_ecube(data, save_dir, voltsperbit):
     '''
     Create a binary file with eCube formatting using the given data
 
+    Note:
+        Importantly, the data and the voltsperbit must have reasonable values, otherwise
+        you will get clipping in the saved data. For instance, if the data ranges from 0-1000,
+        then the voltsperbit should be near to 1. Or if the data ranges from 0-0.001, then the
+        voltsperbit can be lower, around 1e-6. 
+
     Args:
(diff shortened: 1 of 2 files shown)


leoscholl:
ok i figured it out, the data just has to have small enough values otherwise it gets clipped when you divide by voltsperbit.

aopy/utils.py
```diff
@@ -73,21 +79,16 @@ def save_test_signal_ecube(data, save_dir, voltsperbit):
...
     filepath = os.path.join(save_dir, filename)
     with open(filepath, 'wb') as f:
-        for _ in range(8):
-            f.write(np.byte(1)) # 8 byte timestamp
-        for t in range(intdata.shape[0]):
-            for ch in range(intdata.shape[1]):
-                f.write(np.byte(intdata[t,ch]))
-                f.write(np.byte(intdata[t,ch] >> 8))
+        f.write(timestamp)
+        f.write(flatdata.tobytes())
```
zaicheek, line 91: i did not know about this function, thank you.

zaicheek (approved):
docstrings in order, learned a new function for how to wrap data into bytes

leoscholl:
i still can't figure out why the images are slightly different in the test, causing the failure. it also seems to depend on who is running the test because it passes on my machine but not here.

leoscholl:
changes made elsewhere

PR #187, opened 2021-11-08
Add dataset classes mike nolan

m-nolan:
Adds ECoG data reader functions to data.py

load_ecog_clfp_data() reads in clfp files from goose_wireless dataset. Will work on ECoG or LFP data. Must have experiment.json files in local directory.


leoscholl:
this would be a great addition to aopy! i suggest adding "goose" to the names of some of these functions, or something similar to differentiate them from the ecube data loading functions.

question about the DataFile() class: could we get the same functionality just by converting the data into HDF format? so we can read portions of the data without loading the whole thing into memory?

pavi-rajes:
@m-nolan Thanks for sharing the code in a timely manner for me to use it. I think we need to include tests for the function and update docstrings before we can merge. 

Could you update docstrings for the functions ? @leoscholl what is the difference between these loading functions and our ECoG loading functions? If this is specific for Goose dataset, I recommend a separate goose_data.py script instead. Let me know what you think.

m-nolan:
I'll add docstrings to functions I've added.

As for the DataFile() class, it would definitely work better if simply reading from an HDF5 file instead of parsing/reparsing the structure of binary files every time you recreate the class interface. I haven't created HDF5 versions of the goose_wireless data, but I've been meaning to do that for some time. Having the data in compressed array formats would also enable much, much more flexible pytorch dataset/dataloader creation methods than what I'm currently using.

What is our current data conditioning and preprocessing pipeline?

leoscholl:
#181 is the beginning of our preprocessing pipeline. not much else has been written. we have some functions for filtering, band power estimates, spike detection, binning. but no formal pipeline to string those things together.

m-nolan:
@pavi-rajes @leoscholl 

I've added docstrings and resolved the merge conflict. There are some package import overlaps, but the code should run without interference.

leoscholl (commented):
what's the difference between `read_lfp()` and `load_ecog_clfp_data()`?

m-nolan:
Pavi wrote some passing tests for this. Can you review and approve, @leoscholl ?

Thanks!

m-nolan:
@leoscholl
`read_lfp` is a more updated version of `load_ecog_clfp_data` using the `xarray` interface and that's suited to loading more general `file_name.*lfp` file formats than just `*.clfp`.

leoscholl:
thanks for adding the data filtering / quality / outlier functions! they are super helpful broadly, i don't think they belong in this `aopy.data.peslab` subdirectory. would you mind taking them out of here and making a new branch and pull request for those functions?

leoscholl:
do you have a small (<100MB) test dataset that you could upload so we can write some short tests to demo the lfp loading code?

m-nolan:
@leoscholl 

- Moved data quality check code to `/aopy/preproc/quality.py`
- Refactored other `peslab` functions to `utils`
- corrected import calls

The `quality` code currently uses a homebrewed multitaper spectrogram generation function. There is already one of these written in `precondition` for individual PSDs. I'll rewrite the `preproc.quality.mt_sgram` function to use the nitime-based MTPSD estimator.

leoscholl:
cool! the peslab stuff looks good, only exception is the docs will not be complete after refactoring. can you add `peslab` section to `docs/source/data.rst`? see `preproc.rst` as an example. also while you're at it can you fix the `tqdm` import in `docs/source/conf.py` -- see #289 

for the data quality functions, it would be really helpful if we had some tests or examples. would you mind moving quality.py to a different pull request so we can merge all the peslab code now?

m-nolan:
@leoscholl I would mind doing that because (1) the peslab data readers depend on the quality check functions and (2) I don't know how to split pull requests like that.

leoscholl:
Got it. Added some basic tests for the quality check functions and swapped the data axes around to match convention. Could you update the peslab code to follow the same rules? And there are a few errors coming from quality.py but I think they are due to the window for spectrogram being too big for the test data.

m-nolan:
Implemented changes and test alterations made to reflect data array shape changes. Should pass on server.

leoscholl (approved):

PR #189, opened 2021-11-11
Adding task_Codes to a param file

pavi-rajes:
Changes include: 
1) Added task_code.yaml : this is a params file that contains task codes. Please check if the task codes are correct. 
2) Functions to read and write params file in .yaml format. 
3) Included tests for those functions in test_data.py

YAML stands for Yet ANother Markup Language. It is mostly used to save config files. I am saving task codes in this format since it has fixed parameters.


rcanfield2 (approved):
Looks good to me.

PR #191, opened 2021-11-11
add spike classification wrapper

rcanfield2:


rcanfield2:
Addressed issue #210 

Fixed analysis.find_outliers. The mu in the standard deviation calculation was using the mean of the distances instead of 0. It should use 0 since the kmeans centroid is already at the mean of the data points.

aopy/analysis.py
```diff
@@ -67,6 +71,39 @@ def factor_analysis_dimensionality_score(data_in, dimensions, nfold, maxiter=100
 
     return log_likelihood_score, iterations_required
 
+def get_pca_dimensions(data, max_dims=None, VAF=0.9, project_data=False):
+    """
+    Use PCA to estimate the dimensionality required to account for the variance in the given data. If requested it also projects the data onto those dimensions.
+    
+    Args:
+        data (nt, nch): time series data where each channel is considered a 'feature' (nt=n_samples, nch=n_features)
+        max_dims (int): (default None) the maximum number of dimensions
```
leoscholl, line 80: i don't quite get the point of this argument. can you put a bit more detail into the description? or an example?

aopy/analysis.py
```diff
@@ -67,6 +71,39 @@ def factor_analysis_dimensionality_score(data_in, dimensions, nfold, maxiter=100
...
+            | **projected_data (nt, ndims):** Data projected onto the dimensions required to explain the input variance fraction. If the input 'project_data=False', the function will return 'projected_data=None'
+    """
+    pca = PCA()
+    pca.fit(data)
+    explained_variance = pca.explained_variance_ratio_
+    total_explained_variance = np.cumsum(explained_variance)
+    if max_dims is None:
+        num_dims = np.min(np.where(total_explained_variance>VAF)[0])+1
+    else:
+        num_dims = max_dims
```
leoscholl, line 98: seems to me like you would still want to do a comparison between the total_explained_variance, but just take the minimum of num_dims and max_dims

not sure. but this seems a bit pointless to just return the input argument if it's supplied

aopy/analysis.py
```diff
@@ -202,6 +239,88 @@ def calc_success_rate(events, start_events=[b"TARGET_ON"], end_events=[b"REWARD"
...
+    Units with lower spike width are considered inhibitory cells (label: 0) and higher spike width are considered excitatory cells (label: 1)
+    The pulse width is defined as the time between the waveform trough to the waveform peak. (trough-to-peak time)
+    Assumes all waveforms are recorded for the same number of time points. 
+
+    This function conducts the following processing steps:
+
+        | **1. ** For each unit, project each waveform into the top PCs. Number of PCs determined by 'pca_varthresh'
+        | **2. ** For each unit, remove outlier spikes. Outlier threhsold determined by 'std_threshold'. If the number of waveforms is less than 'min_wf', no waveforms are removed.
+        | **3. ** For each unit, average remaining waveforms.
+        | **4. ** For each unit, calculate spike width using a local polynomail interpolation.
```
leoscholl, line 257:
```suggestion
        | **4. ** For each unit, calculate spike width using a local polynomial interpolation.
```

aopy/analysis.py
```diff
@@ -202,6 +239,88 @@ def calc_success_rate(events, start_events=[b"TARGET_ON"], end_events=[b"REWARD"
...
+        | **5. ** Use a gaussian mixture model to classify all units
+
+    Args:
+        waveform_data (nunit long list of (nt x nwaveforms) arrays): Waveforms of each unit. Each element of the list is a 2D array for each unit. Each 2D array contains the timeseries of all recorded waveforms for a given unit.
+        samplerate (float): sampling rate of the points in each waveform. 
+        std_threshold (float): For outlier removal. The maximum number of standard deviations (in PC space) away from the mean a given waveform is allowed to be. Defaults to 3
+        pca_varthresh (float): Variance threshold for determining the number of dimensions to project spiking data onto. Defaults to 0.75.
+        min_wf_samples (int): Minimum number of waveform samples required to perform outlier detection.
+        nspikesamples (None or int): Optional, defaults to None. Specificy how many waveforms to use in classification. Randomly selects the set amount of waveforms for each unit. If there are more waveforms asked for than recorded spikes all of the recorded spikes will be used.
```
leoscholl, line 266: a bit unclear what this parameter does. if it's none, it randomly selects an amount? or if it's an integer then it selects that many randomly and if it's none it takes all of them?

tests/test_analysis.py
```diff
@@ -104,6 +132,7 @@ def test_get_pca_dimensions(self):
         single_num_dims = 1
 
         VAF, num_dims, proj_data = aopy.analysis.get_pca_dimensions(single_dim_data)
```
leoscholl, line 134: could you add a call to get_pca_dimensions() with all the optional arguments to prove that they actually do something now

leoscholl (approved):
still not convinced by num_dims, but nobody is using it anyway i guess

aopy/analysis.py
rcanfield2: The intent was that the function could subselect 'nspikesamples' number of waveforms to average. I think it makes more sense to pull this out of the function and have the user subselect their samples before inputting them into this wrapper.

rcanfield2: oh wait, i did pull it out and just forgot to remove it from the documentation

PR #196, opened 2021-11-15
Leo animation

leoscholl:
i fixed a bug that was causing the framerate of saved animations to always be 5 fps
and added a function to animate spatial maps


aopy/visualization.py
```diff
@@ -352,6 +350,59 @@ def draw(num):
...
+            saveanim(ani, write_dir, filename)
+
+        .. raw:: html
+
+            <video controls src="_static/spatial_map_animation.mp4"></video>
+
+    Args:
+        data_map (nt): array of 2d maps
+        x (list): list of x positions
+        y (list): list of y positions
```
rcanfield2, line 382: According to the example data_map should be an nt long list of datamaps (not sure what dimension these are). And x and y should be numpy arrays.

rcanfield2 (approved):
Looks good. As a separate issue, we should probably make an example notebooks showing how to call animation and save results. Maybe we (I) can clean the laser test heatmap jupyter notebook up and use that.

PR #202, opened 2021-11-21
clean curvefitting

rcanfield2:


aopy/analysis.py
```diff
@@ -111,9 +113,18 @@ def get_preferred_direction(b1, b2):
     .. math:: 
         
         arctan(\\frac{b_1^2}{b_2^2})
-        
+    
+    Returns:
+        float: Preferred direction of the curve fit in radians
```
pavi-rajes, line 118: This function is not returning preferred direction in radians anymore. Any reason to covert pd to degrees here instead of converting outside the function?

pavi-rajes (commented):
Code changes and test functions look good. But I think get_pd() should return pd in radians and conversion to degrees using np.rad2deg() could be suggested in a note in the docstrings. 

My reason being PD is often talked about in terms of radians and common models use cosine tuning fits of 2*pi again in radians. It is also easy to plot in polar plots (see: matplotlib.pyplot.polar()) as well.

pavi-rajes (approved):
Changes look good to me.

PR #205, opened 2021-11-28
Ryan add misc functions

rcanfield2:


aopy/utils.py
```diff
@@ -316,3 +316,33 @@ def max_repeated_nans(a):
     else:
         idx = np.nonzero(mask[1:] != mask[:-1])[0]
         return (idx[1::2] - idx[::2]).max()
+
+'''
+Matrix/array manipulation
+'''
+
+def reorder_dataidx(data_metric):
```
m-nolan, line 324: There are numpy builtins that handle this: [np.sort](https://numpy.org/doc/stable/reference/generated/numpy.sort.html) and [np.argsort](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html).

I don't know if there's a speed-up or memory load benefit with yours or the numpy version. I don't think there's a problem here, but consider using the package built-ins if possible.

m-nolan (approved):
Looks good! Curious about the array sorting method though.

aopy/utils.py
rcanfield2: Great thanks, I must have missed argsort. I'll just remove this function

PR #206, opened 2021-12-09
Get instantaneous target direction.

rcanfield2:
In progress, still need to add other functions and tests

### DIFF
--- a/aopy/postproc.py
+++ b/aopy/postproc.py
@@ -136,71 +136,81 @@ def get_trial_targets(trials, targets):
         trial_targets[trial].append(targets[idx])
     return trial_targets
 
-def get_target_dir_from_cursor(cursor_pos, target_pos):
+def get_relative_point_location(ref_point_pos, new_point_pos):
     '''
-    This function calculates the instantaneous target direction from the current cursor position. Assumes angle 0 starts from the direciton of vector [1, 0]
-    This function is specific to the 2D case at the moment.
+    This function calculates the relative location (angle and position) of a point compared to a reference point.
+    Assumes angle 0 starts from the direciton of vector [1, 0].
+    This function is specific to the 2D case at the moment but can handle a single point entry or multiple.
 
     Args:
-        cursor_pos (x, y): Current cursor position in spatial coordinates
-        target_pos (x, y): Current target position in spatial coordinates
+        ref_point_pos (nxpts, nypts): Point position of reference point in spatial coordinates
+        new_point_pos (nxpts, nypts): Point position of new point in spatial coordinates
 
     Returns:
         A tuple containing:
-            | **relative_target_angle (float):** Absolute angle between cursor and target position [rad]
-            | **relative_target_pos (float):** Absolute position of the target relative to the cursor
+            | **relative_new_point_angle (float):** Absolute angle between cursor and target position [rad]
+            | **relative_new_point_pos (float):** Absolute position of the target relative to the cursor
     '''
-    relative_target_pos = target_pos - cursor_pos
-    relative_target_angle = np.arctan2(relative_target_pos[1], relative_target_pos[0])
-    if relative_target_angle < 0:
-        relative_target_angle = 2*np.pi + relative_target_angle
+    # Handle single point case
+    relative_new_point_pos = new_point_pos - ref_point_pos
 
-    return relative_target_angle, relative_target_pos
+    if len(ref_point_pos.shape) == 1 or relative_new_point_pos.shape[1] == 1:
+        relative_new_point_angle = np.arctan2(relative_new_point_pos[1], relative_new_point_pos[0])
+        if relative_new_point_angle < 0:
+            relative_new_point_angle = 2*np.pi + relative_new_point_angle
+    
+    # Handle multi-point case
+    else:
+        relative_new_point_angle = np.arctan2(relative_new_point_pos[:,1], relative_new_point_pos[:,0])
+        relative_new_point_angle_mask = relative_new_point_angle < 0
+        relative_new_point_angle[relative_new_point_angle_mask] = 2*np.pi + relative_new_point_angle[relative_new_point_angle_mask]
 
-def get_inst_target_dir(trialaligned_xpos, trialaligned_ypos, targetpospertrial):
+    return relative_new_point_angle, relative_new_point_pos
+
+def get_inst_target_dir(trial_aligned_pos, targetpospertrial):
     '''
     This function calculates the instantaneous direction from the cursor to the target at each time point across each trial.
-    This function is specific to the 2D case at the moment.
+    This function is specific to the 2D case at the moment with X coordinates being the horizontal dimension and Y coordinates being the vertical dimension. 
 
     Args:
-         trialaligned_xpos (ntime, ntrials): X-position of the cursor
-         trialaligned_ypos (ntime, ntrials): Y-position of the cursor
-         targetpospertrial (ntrials, 2): X and Y pos of target
+         trial_aligned_pos (ntime, ntrials, 2): Position of the cursor in X and Y coordinates.trial_aligned_pos[:,:,0] corresponds to the X cursor positions and trial_aligned_pos[:,:,1] to the Y cursor positions.
+         targetpospertrial (ntrials, 2): X and Y pos of target.
 
     Returns:
         (ntime, ntrials): Array including instantaneous direction to the target from the cursor [rad]
 
     '''
-    ntime = trialaligned_xpos.shape[0]
-    ntrials = trialaligned_xpos.shape[1]
+    ntime = trial_aligned_pos.shape[0]
+    ntrials = trial_aligned_pos.shape[1]
     inst_target_dir = np.zeros((ntime, ntrials))*np.nan
 
     for itrial in range(ntrials):
-        for itime in range(ntime):
-            cursor_location = [trialaligned_xpos[itime, itrial], trialaligned_ypos[itime, itrial]]
-            target_location = targetpospertrial[itrial, :]
-            rel_target_angle, _ = get_target_dir_from_cursor(cursor_location, target_location)
-            inst_target_dir[itime, itrial] = rel_target_angle
+        cursor_location = trial_aligned_pos[:, itrial,:]
+        target_location = targetpospertrial[itrial, :]
+        rel_target_angle, _ = get_relative_point_location(cursor_location, target_location)
+        inst_target_dir[:, itrial] = rel_target_angle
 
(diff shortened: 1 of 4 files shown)


aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+        A tuple containing:
+            | **relative_target_angle (float):** Absolute angle between cursor and target position [rad]
+            | **relative_target_pos (float):** Absolute position of the target relative to the cursor
+    '''
+    relative_target_pos = target_pos - cursor_pos
+    relative_target_angle = np.arctan2(relative_target_pos[1], relative_target_pos[0])
+    if relative_target_angle < 0:
+        relative_target_angle = 2*np.pi + relative_target_angle
+
+    return relative_target_angle, relative_target_pos
```
leoscholl, line 158: can we just call this something like `angle_between_points(p1, p2)` to make it more general?

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+
+def get_inst_target_dir(trialaligned_xpos, trialaligned_ypos, targetpospertrial):
+    '''
+    This function calculates the instantaneous direction from the cursor to the target at each time point across each trial.
+    This function is specific to the 2D case at the moment.
+
+    Args:
+         trialaligned_xpos (ntime, ntrials): X-position of the cursor
+         trialaligned_ypos (ntime, ntrials): Y-position of the cursor
+         targetpospertrial (ntrials, 2): X and Y pos of target
```
leoscholl, line 168: more underscores would make these names easier to read

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+    ntime = trialaligned_xpos.shape[0]
+    ntrials = trialaligned_xpos.shape[1]
+    inst_target_dir = np.zeros((ntime, ntrials))*np.nan
+
+    for itrial in range(ntrials):
+        for itime in range(ntime):
+            cursor_location = [trialaligned_xpos[itime, itrial], trialaligned_ypos[itime, itrial]]
+            target_location = targetpospertrial[itrial, :]
+            rel_target_angle, _ = get_target_dir_from_cursor(cursor_location, target_location)
+            inst_target_dir[itime, itrial] = rel_target_angle
```
leoscholl, line 183: it would be nice if you could vectorize this a bit more. np.arctan2 works element-wise, so you could input all the timepoints for a single trial at the same time into `get_target_dir_from_cursor()`

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+
+def mean_fr_inst_dir(data, cursorxpos, cursorypos, targetpos, data_binwidth, targetloc_binwidth, data_samplerate, cursor_samplerate):
+    '''
+    This function takes trial aligned neural data, cursor position, and target position and calculates
+    the mean firing rate per target location. 
+
+    Args:
+        data (ntime, nunit, ntrial): Trial aligned data
+        cursorxpos (ntime, ntrial): Trial aligned cursor position
+        cursorypos (ntime, ntrial): Trial aligned cursor position
```
leoscholl, line 195: not sure why you want to keep x and y separate rather than have a (ntime, ntrial, 2) or (ntime, 2, ntrial) array

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+            target_location = targetpospertrial[itrial, :]
+            rel_target_angle, _ = get_target_dir_from_cursor(cursor_location, target_location)
+            inst_target_dir[itime, itrial] = rel_target_angle
+
+    return inst_target_dir
+
+def mean_fr_inst_dir(data, cursorxpos, cursorypos, targetpos, data_binwidth, targetloc_binwidth, data_samplerate, cursor_samplerate):
+    '''
+    This function takes trial aligned neural data, cursor position, and target position and calculates
+    the mean firing rate per target location.
```
leoscholl, line 190: more explanation of the "per target location" part would be good. it's not immediately clear how the averaging works.

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+        (nunit, ndirection): Average firing rate per unit per direction bin.
+        
+    '''
+    # Check that binwidths are not lower than allowed by sampling rate
+    max_cursor_binwidth = 1/cursor_samplerate #[s/sample]
+    if data_binwidth < max_cursor_binwidth:
+        data_binwidth = max_cursor_binwidth
+    
+    ndatatime, nunit, ntrial = data.shape
+    ndirection = 360//targetloc_binwidth
```
leoscholl, line 212: today i learned how to do integer division in python 3

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+    This function takes trial aligned neural data, cursor position, and target position and calculates
+    the mean firing rate per target location. 
+
+    Args:
+        data (ntime, nunit, ntrial): Trial aligned data
+        cursorxpos (ntime, ntrial): Trial aligned cursor position
+        cursorypos (ntime, ntrial): Trial aligned cursor position
+        targetpos (ntrial, 2): Target position for each trial. First column is x target position, second column is y target position.
+        data_binwidth (float): Bin size for neural data and cursor position. Can not be smaller than allowed by cursor position sampling rate.
+        targetloc_binwidth (float): Bin size for lumping target positions [deg]
```
leoscholl, line 198: not sure i'd call this a binwidth, more like number of directions. also it might be nice to have this be an optional argument, and if you leave it blank it picks the number of directions to be the same as the number of unique targets

leoscholl: nevermind about the optional argument part. i get it now, since it's instantaneous there are always infinite directions.. doesn't matter how many targets there are

aopy/postproc.py
leoscholl: maybe you can include a reference to `get_inst_target_dir()` to make it very clear what the underlying computation is

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+    if 360%targetloc_binwidth != 0:
+        warnings.warn("Target location bins are not evenly spaced. Please choose a bin width that evenly divides into 360 degrees")
+
+    # Bin neural data and cursor pos for each trial to make them the same samplingrate
+    binned_data = np.zeros((nbins, nunit, ntrial))*np.nan
+    for itrial in range(ntrial):
+        binned_data[:,:,itrial] = precondition.bin_spikes(data[:,:,itrial], data_samplerate, data_binwidth)
+
+    binned_cursorxpos = precondition.bin_spikes(cursorxpos, cursor_samplerate, data_binwidth)
+    binned_cursorypos = precondition.bin_spikes(cursorypos, cursor_samplerate, data_binwidth)
```
leoscholl, line 225: hmmmmmm

i guess we should rename bin_spikes to something else, huh

leoscholl: actually are you sure this works? seems to me summing cursor positions together is a bad idea

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+
+    # Get instantaneous target location for each cursor pos (ntime, ntrial)
+    #print(binned_cursorxpos.shape, binned_cursorypos.shape, targetpos.shape)
+    inst_target_dir = get_inst_target_dir(binned_cursorxpos, binned_cursorypos, targetpos)
+    target_binid = 1 + (inst_target_dir-(np.deg2rad(targetloc_binwidth)/2))//np.deg2rad(targetloc_binwidth)
```
leoscholl, line 230: this line could be easier to read. maybe something like:
```suggestion
    # a comment here would be nice
    target_width_rad = np.deg2rad(targetloc_binwidth)
    target_binid = 1 + (inst_target_dir - (target_width_rad / 2)) // target_width_rad
```

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+    target_binid[target_binid==ndirection] = 0
```
leoscholl, line 231:
```suggestion
    target_binid[target_binid==ndirection] = 0 # to combine the first and last bins
```

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+
+    # Average data and place into correct points in array
+    mean_dir_fr = np.zeros((nunit, ndirection))
+    for iunit in range(nunit):
```
leoscholl, line 235: i think this loop over units is unnecessary, you should be able to average across bins for all units at the same time

aopy/postproc.py
rcanfield2: My understanding was that it was averaging the cursor position for each bin

aopy/postproc.py
rcanfield2: Hmmm i agree but i can't quite figure out how to do this since i'm masking binned_data with target_binid creating a 1D array. Therefore, I'm having trouble figuring out how to separate each unit

aopy/postproc.py
leoscholl: i guess so? its summing the cursor positions then dividing by the bin width, so won't it be off by some factor of time to the arithmetic average cursor position? but i suppose that doesn't matter since the relative direction will be the same between any two points.

aopy/postproc.py
```diff
@@ -136,6 +136,109 @@ def get_trial_targets(trials, targets):
...
+        for idir in range(ndirection):
+            temp_data = binned_data[:,iunit,:]
+            mean_dir_fr[iunit, idir] = np.mean(temp_data[target_binid==idir])
```
leoscholl, line 248: have you tried something like this:
```suggestion
    for idir in range(ndirection):
        mean_dir_fr[:, idir] = np.mean(binned_data[target_binid==idir,:,:], axis=(0,2))
```

leoscholl (approved):
looks good!

aopy/postproc.py
rcanfield2: oh i see, yes, let me think about this some more.

rcanfield2:
You were right about the error induced by using bin_spikes for averaging position. To compensate i just multiplied the output by bin_width and divided by samples/bin. This led me to a bug in bin_spikes where i divided the output by the wrong variable to convert from spikes/bin to spikes/s.

PR #209, opened 2021-12-19
Add function to set colormap

rcanfield2:
As I mentioned in the issue, I'm not sure if we need this or should just use sns.set_palette()

### DIFF
--- a/aopy/visualization.py
+++ b/aopy/visualization.py
@@ -892,3 +892,18 @@ def plot_boxplots(data, plt_xaxis, trendline=True, facecolor=[0.5, 0.5, 0.5], li
             boxprops=dict(facecolor=facecolor, color=linecolor), capprops=dict(color=linecolor),
             whiskerprops=dict(color=linecolor), flierprops=dict(color=facecolor, markeredgecolor=facecolor),
             medianprops=dict(color=linecolor))
+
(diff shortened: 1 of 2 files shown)


aopy/visualization.py
```diff
@@ -892,3 +892,18 @@ def plot_boxplots(data, plt_xaxis, trendline=True, facecolor=[0.5, 0.5, 0.5], li
             boxprops=dict(facecolor=facecolor, color=linecolor), capprops=dict(color=linecolor),
             whiskerprops=dict(color=linecolor), flierprops=dict(color=facecolor, markeredgecolor=facecolor),
             medianprops=dict(color=linecolor))
+
+def get_plotcolors(nitems, colormap="viridis"):
```
leoscholl, line 896: maybe set_plotcolors()? i was imagining a function that returns a list of colors, for example to be used in color_trajectories(), above. but we can change things around so that function just grabs colors from the current pallette.

leoscholl:
leaving a comment here in case anyone finds this useful: 
```
import matplotlib.pyplot as plt
import matplotlib.colors as mcolors

def get_plot_colors(n_colors, colormap_name='viridis'):
    cmap = plt.get_cmap(colormap_name)
    values = np.linspace(0, 1, n_colors)
    return [cmap(v) for v in values]

```

Issue #208, opened 2021-12-18
butterworth filter function only works with bandpass


leoscholl:
https://github.com/aolabNeuro/analyze/blob/0b6947d3dae7c619305450cc170a7f59829faf98/aopy/precondition.py#L64

the `or 'bandstop'` makes this line always True

m-nolan:
@leoscholl @pavi-rajes Why does the `or 'bandstop'` break this? That block won't run unless `filter_type` is either `bandpass` or `bandstop`. The former is `filter_type`'s default value, but if neither are true then the code skips the block.

Maybe using a `if filter_type in ['list', 'of', 'filter', 'types]:` construction would fix this?

leoscholl:
`if filter_type in ['list', 'of', 'filter', 'types']:` would be good, yes. 

in `if filter_type == 'bandpass' or 'bandstop'`, the `or` combines two boolean expressions: `filter_type == 'bandpass'` on the left, which will be `False` by default, and `'bandstop'` on the right, which if you call `bool('bandstop')` you will see equates to `True` always. 

so the other way to fix this would be `if filter_type == 'bandpass' or filter_type == 'bandstop':`

leoscholl:
workaround for now is to specify butterworth order

PR #216, opened 2022-01-07
Issue 208

m-nolan:
#208 resolves logic issues in filter_type control flow


m-nolan:
Looks like I accidentally swept a few other unresolved code changes into this one. I'll see if I can separate them, but it passes precondition tests.

leoscholl (approved):

PR #220, opened 2022-01-08
add nan handling for a couple functions

rcanfield2:


aopy/analysis.py
```diff
@@ -214,11 +215,22 @@ def run_tuningcurve_fit(mean_fr, targets):
...
+            pd[iunit] = get_preferred_direction(params[0], params[1])
+
+        # If this doesn't work, check if fit_with_nans is true. It it is remove nans and fit
+        elif fit_with_nans:
+            nonnanidx = ~np.isnan(mean_fr[iunit,:])
+            params, _ = curve_fit(curve_fitting_func, targets[nonnanidx], mean_fr[iunit,nonnanidx])
+            fit_params[iunit,:] = params
+
+            md[iunit] = get_modulation_depth(params[0], params[1])
+            pd[iunit] = get_preferred_direction(params[0], params[1])
```
leoscholl, line 233: could you move these two lines out one indent so you don't have to copy them twice?

leoscholl (approved):

aopy/analysis.py
rcanfield2: Maybe i should restructure the logic. Right now, if the input arguments don't fit into one of those if statements no fitting is performed leaving the value as nan. If i pull out the fitting there i need more logic to only call it in specific conditions, which i'm not sure is more intuitive than the way it is now. For now, I'll merge these commits but I can update in the future.

PR #224, opened 2022-01-12
make a preprocessing pipeline

leoscholl:
in progress


leoscholl:
to do: 
- speedup lfp filtering
- write function to copy data from one location to another
- write wrapper to call one function after another
- use wrapper to copy raw ecube data, then turn it into an hdf5 file, then filter that into lfp into the same file

leoscholl:
i've been using this function to copy the broadband data from binary files into an hdf dataset. can you help me figure out a better way to set the size of chunks to load at once? currently it's in seconds, but that means if you have a lot of channels it reads bigger chunks at once
'''
def proc_broadband(data_dir, files, result_dir, result_filename, overwrite=False, batchsize=1.):
    '''
    Process broadband data:
        Loads 'ecube' headstage data and metadata
    Saves broadband data into the HDF datasets:
        broadband_data (nt, nch)
        broadband_metadata (dict)
    
    Args:
        data_dir (str): where the data files are located
        files (dict): dictionary of filenames indexed by system
        result_filename (str): where to store the processed result
        overwrite (bool, optional): whether to remove existing processed files if they exist
        batchsize (float, optional): time in seconds for each batch to be processed

    Returns:
        None
    '''  
    # Check if a processed file already exists
    filepath = os.path.join(result_dir, result_filename)
    if not overwrite and os.path.exists(filepath):
        contents = aopy.data.get_hdf_dictionary(result_dir, result_filename)
        if "lfp_data" in contents:
            print("File {} already preprocessed, doing nothing.".format(result_filename))
            return
    elif os.path.exists(filepath):
        os.remove(filepath) # maybe bad, since it deletes everything, not just lfp_data

    # Preprocess neural data into lfp
    if 'ecube' in files:
        data_path = os.path.join(data_dir, files['ecube'])
        metadata = aopy.data.load_ecube_metadata(data_path, 'Headstages')
        samplerate = metadata['samplerate']
        chunksize = int(batchsize * samplerate)

        # Create an hdf dataset
        result_filepath = os.path.join(result_dir, result_filename)
        hdf = h5py.File(result_filepath, 'a')
        dset = hdf.create_dataset('broadband_data', (metadata['n_samples'], metadata['n_channels']), dtype='int16')

        # Write broadband data directly into the hdf file
        n_samples = 0
        for broadband_chunk in aopy.data.load_ecube_data_chunked(data_path, 'Headstages', chunksize=chunksize):
            chunk_len = broadband_chunk.shape[0]
            dset[n_samples:n_samples+chunk_len,:] = broadband_chunk
            n_samples += chunk_len
        hdf.close()

        # Append the broadband metadata to the file
        aopy.data.save_hdf(result_dir, result_filename, metadata, "/broadband_metadata", append=True)
'''

leoscholl:
i've added proc_broadband()

aopy/preproc/__init__.py
```diff
@@ -3,12 +3,57 @@
...
+from ..data import load_ecube_data_chunked, load_ecube_metadata, proc_ecube_data, save_hdf, load_hdf_group, get_hdf_dictionary
 import os
+import h5py
+from glob import glob
 
 '''
 proc_* wrappers
 '''
+def proc_day(data_dir, result_dir, overwrite=False, save_res=True):
+    files = glob(os.path.join(data_dir,'*.bin'))
```
leoscholl, line 15: Better if files is an input to this function since it will come from database code that isn’t included here.

leoscholl:
@m-nolan can you add a docstring to proc_day(), otherwise i think this is done. can you give it a once-over and make sure everything works before we merge? there's still a couple things to do before we have a "complete" preprocessing pipeline, e.g. 
- copy data from one location to another
- use broadband hdf data to filter into lfp rather than regrabbing the ecube data
but i think these can be future enhancements. a lot of code here could be useful now.

m-nolan (approved):

PR #227, opened 2022-01-15
Make proc_eyetracking return eye data

pqz317:
proc_eyetracking can now return eye data, add optional bool to control saving the data to filesystem. 
This will allow for quickly recalculating eye tracking data without touching filesystem. 

Testing: ran updated unit test


leoscholl (approved):
I like this idea a lot, we should apply the same pattern to all the proc_() functions I think

PR #228, opened 2022-01-17
update run_tuningcurve_fit

rcanfield2:
I added a provision in analysis.run_tuningcurve_fit to return a nan for units that don't have a minimum number of directional data points.

i.e. if a unit has nan values for each target direction, the function will return a nan.


leoscholl (approved):
look good

PR #233, opened 2022-01-19
consider only valid indices for calibration

pqz317:
Not sure if this fix is necessary, or if its the right fix: 
When running eye calibration for an older session, te2337 from [Si Jia's example script](http://aolab.wanprc.washington.edu:8888/notebooks/sijia/eyetracking_double_eye_session2337.ipynb), I was getting this error:
```
/Users/patrick/venv/bin/python /Users/patrick/analyze/aopy/eye_track_test_script.py
eye channel definitions do not exist, use eye channels [9, 8, 10, 11] 
Using 208 cursor x,y positions to calibrate eye tracking data
Traceback (most recent call last):
  File "/Users/patrick/analyze/aopy/eye_track_test_script.py", line 31, in <module>
    save_res=False
  File "/Users/patrick/analyze/aopy/preproc.py", line 1525, in proc_eyetracking
    event_cycles, event_times, event_codes, debug=debug, return_datapoints=True, **kwargs)
  File "/Users/patrick/analyze/aopy/preproc.py", line 562, in calc_eye_calibration
    eye_data_aligned = eye_data[sample_eye_enter_target,:]
IndexError: index -9223372036854775808 is out of bounds for axis 0 with size 17093298
Process finished with exit code 1
```
Where the index -9223372036854775808 corresponds to MIN_INT of int64. This seems to be due to NaN values existing in the `align_times`. This fix just filters those NaN values out. 
Alternatively Amy mentioned that some of the older trial data got corrupted so if this issue is due to data corruption maybe we don't need this check?


leoscholl (approved):
makes sense. opening issue #234 to look into why this is happening but your fix looks good for now.

Issue #231, opened 2022-01-18
Return data and metadata in proc_() functions


leoscholl:
See https://github.com/aolabNeuro/analyze/pull/227#pullrequestreview-854852350_

Similar return statements could be added to other preproc functions

sijia66:
I am taking some notes to make this issue easier to resolve. 

for proc_exp: 
1. add save_res as an optional argument 
2. output bmi3d_data, bmi3d_metadata

for proc_mocap:
1.  add save_res as an optional argument 
2. output  optitrack_data and optitrack_metadata

for proc_lfp:
1.  add save_res as an optional argument 
2. output lfp_metadata, lfp_data

leoscholl:
done for proc_exp()

leoscholl:
need to think a bit more about whether this is actually useful. Currently the preprocessing functions always have to do work even if overwrite=False and the file already exists, since they need to return the data and metadata.

PR #235, opened 2022-01-25
move bmi3d preproc code into submodule

leoscholl:
also added save_res to proc_exp() to address #231


m-nolan (approved):
Clean refactor, passes tests.

PR #237, opened 2022-01-31
Update __init__.py

leoscholl:


leoscholl:
can you check that this works?

pqz317 (approved):
Looks like this fix is working, I was successfully able to run one preproc functions on one session of affi data in: http://aolab.wanprc.washington.edu:8888/notebooks/patrick/notebooks/Preprocess%20Affi.ipynb
Generating `/data/preprocessed/affi/preprocessed_te3820.hdf`

PR #241, opened 2022-02-08
add mean and max ERP calculations

leoscholl:


aopy/analysis.py
```diff
@@ -817,6 +817,75 @@ def calc_sem(data, axis=None):
...
+    '''
+    Calculates the mean (across trials) event-related potential (ERP) for the given timeseries data.
+
+    Args:
+        data (nt, nch): timeseries data across channels
+        event_times (ntrial): list of event times
+        time_before (float): number of seconds to include before each event
+        time_after (float): number of seconds to include after each event
+        samplerate (float): sampling rate of the data
+        subtract_baseline (bool, optional): if True, subtract the mean of the aligned data during
```
aolabNeuro, line 831: I'd suggest adding an optional input argument that allows you to specify the window of time to use for computing the baseline. Current implementation could be the default, but would increase flexibility to let it be set.

aopy/analysis.py
```diff
@@ -817,6 +817,75 @@ def calc_sem(data, axis=None):
...
+    for the given timeseries data.
+
+    Args:
+        data (nt, nch): timeseries data across channels
+        event_times (ntrial): list of event times
+        time_before (float): number of seconds to include before each event
+        time_after (float): number of seconds to include after each event
+        samplerate (float): sampling rate of the data
+        subtract_baseline (bool, optional): if True, subtract the mean of the aligned data during
+            the time_before period preceding each event. Must supply a positive time_before. Default True
```
aolabNeuro, line 875: I would add an optional argument to search for the maximum within a particular time-range. This can help a lot with data that has more complex dynamics, and make calculations more robust if you change parameters (e.g. load more total data where you end up getting peaks later on for reasons not related to stimulus you're aligning to)

tests/test_analysis.py
```diff
@@ -318,6 +318,47 @@ def test_calc_sem(self):
...
+        data = np.zeros(((1+nevents)*samplerate, nch))
+
+        # Make the data zero everywhere except for one sample after each event time
+        data[[int(t)+1 for t in event_times*samplerate],0] = 1
+        data[[int(t)+1 for t in event_times*samplerate],1] = 2
+
+        max_erp = aopy.analysis.calc_max_erp(data, event_times, 0.1, 0.1, samplerate)
+        self.assertTrue(max_erp[0] > 0.5) 
+        self.assertTrue(max_erp[1] > 1.)
+        # not sure why the max erp isn't exactly 1 (or 2 for ch2), it seems to get spread out over a couple samples
```
aolabNeuro, line 360: that seems concerning. Could you show some plots of the test data (raw + trial-aligned) so we can go over this? Seems like something to sort out.

aolabNeuro (commented):
Looks pretty good. I made a few comments for suggested feature additions to add in future versions. I'd also like to go over the test function not giving what you expect to nail that issue down--that's top priority before we merge.

aopy/analysis.py
leoscholl: added baseline_window

aopy/analysis.py
leoscholl: added max_search_window

tests/test_analysis.py
leoscholl: fixed the rounding error that was causing this. probably not a big deal at 1khz or 25khz sampled data to be off by one sample, but anyway good that it works as expected now.

aopy/analysis.py
aolabNeuro: Looks good!

aopy/analysis.py
aolabNeuro: Looks good

aolabNeuro (approved):
Seems good to go.

PR #245, opened 2022-02-16
made code computing inter spike interval histogram

resatomo:
For plotting, I couldn't plot a figure even if `plot_flag = True`. 
But, I was able to show the figure when doing `plt.show()` in the command line. Is there something wrong?


aopy/analysis.py
```diff
@@ -599,3 +599,58 @@ def calc_freq_domain_amplitude(data, samplerate, rms=False):
...
+    dT = 1/fs
+    hist_bins = np.arange(0, hist_width, bin_width)
+
+    ISI_hist = np.zeros((len(hist_bins)-1, n_unit))
+    for iU in range(n_unit):
+        spike_idx = np.where( data[:,iU] )
+        ISI = np.diff(spike_idx)*dT
+        ISI_hist[:,iU], _ = np.histogram(ISI, hist_bins)
+    
+    hist_bins = hist_bins[:-1] + np.diff(hist_bins)/2 # change hist_bins to be the center of the bin, not the edges
```
m-nolan, line 647: the offset value could be simpler: `hist_bins = hist_bins[:-1] + bin_width/2`

m-nolan (approved):
Looks good to me. Seems like there may be some merge conflicts; clear that up and you're good to go.

PR #246, opened 2022-02-17
fix preproc docs

rcanfield2:


zaicheek (approved):
thank you for adding preproc.py readthedoc functionality!

PR #249, opened 2022-02-19
add accumulated log-likelihood (AccLLR)

rcanfield2:
Based on this paper: https://pesaranlab.org/wp-content/uploads/2014/03/BanerjeeDeanPesaran2010.pdf

I'm a little unsure what the 'delta t' term in equation 5 of the paper is and likely implemented it incorrectly.

I'm unsure what the best way to calculate the AccLLR threshold used for classification and determining the selection time. I can't quite tell how they determined their max AccLLR to threshold.

Still need to add test (not sure what the best way to do this would be)


leoscholl:
for determining the threshold it sounds like we need to write some more code separately that sweeps this function over the thresh_proportion parameter, estimates false alarm rates, and then we can choose a value for thresh_proportion that gives us 0.05 false alarm rate for example. 

to test we could use some real lfp & spiking data? i have lfp data from optogenetics or visually evoked stimuli

rcanfield2:
For the AccLLR threshold I guess I was more confused about how the specific value for maxAccLLR was determined. Is it determined on a trial-trial basis or trial averaged? If we choose a value that gives a 0.05 false alarm rate then do we train it on the 'cond1'/'cond2' datasets and have a separate testing dataset or just determine the false alarm rate on the training set?

Instead of having the user input 1D vectors for 'cond1'/'cond2' should we instead have the user input (npts, ntrials) size datasets to do training on?

Sounds good for the test.

rcanfield2:
Need to add code functionality to invoke Poisson spiking as 0's or 1's

aopy/analysis.py
```diff
@@ -856,3 +854,128 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+    '''
+
+    thresh_props = np.arange(0, 1, step_size)
+    fa_rates = []
+
+    # Define some test data and some training data?
+    cond1 = data_cond1[:,0]
+    cond2 = data_cond2[:,0]
+    data = np.concatenate((data_cond1[:,1:], data_cond2[:,1:]))
+    data_labels = np.concatenate((np.ones(len(data_cond1)-1), 2*np.ones(len(data_cond2)-1)))
```
rcanfield2, line 963: I think we need to have the user input training and test data for each condition.

aopy/analysis.py
```diff
@@ -856,3 +854,128 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+    
+    for tp in thresh_props:
+        accLLR, selection_time_idx = calc_activity_onset_accLLR(data, cond1, cond2, modality, bin_width, thresh_proportion=tp)
```
rcanfield2, line 966: I think to calculate false alarm you input the cond2 data and see how many reach the cond1 (+) threshold.

aopy/analysis.py
```diff
@@ -856,3 +854,128 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+        accLLR_cond1_within_time = np.logical_and(accLLR > 0, selection_time_idx < data.shape[0])
+        n_cond2 = np.count_nonzero(data_labels == 2)
```
rcanfield2, line 968: I'm a little confused what the point of this line is.

rcanfield2:
I think we need to go through the functions and define the sizes we expect to pass through each area

rcanfield2:
Todo:
1. Update accLLR calculation function with a flag to trail average or run on each individual trial
2. Update threshold calculation function based on my comments
3. Write wrapper
4. Write tests with artifical spiking (0,0,0,1,1,1) data and artificial LFP data (0,0,0,1,2,3,4)

aopy/analysis.py
```diff
@@ -856,3 +854,146 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
     linear_fit = reg_fit.coef_[0][0]*xdata.flatten() + reg_fit.intercept_
 
     return linear_fit, linear_fit_score, pcc_all[0], pcc_all[1], reg_fit
+
+def calc_activity_onset_accLLR(data, cond1, cond2, modality, bin_width, thresh_proportion=0.15, trial_average=True):
```
rcanfield2, line 858: Kind of nit-picky but i think the 'trial_average' input arg should be False. At least thats how I plan to use it.

aopy/analysis.py
```diff
@@ -856,3 +854,179 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+    train_data_cond2 = build_data_cond2[train_trial_mask,:]
+    valid_data_cond1 = build_data_cond1[np.logical_not(train_trial_mask),:]
+    valid_data_cond2 = build_data_cond2[np.logical_not(train_trial_mask),:]
+
+    # Input train and validate data to calculate acclllr threshold (transpose data back to input into accllr functions)
+    accllr_thresh = calc_accLLR_threshold(train_data_cond1.T, train_data_cond2.T, valid_data_cond2.T, modality, bin_width, step_size=step_size, false_alarm_prob=false_alarm_prob)
+
+    # Use the calculated threshold to run accllr on the test datasets
+    accllr_cond1, selection_time_cond1 = calc_activity_onset_accLLR(test_data_cond1, cond1, cond2, modality, bin_width, thresh_proportion=accllr_thresh, trial_average=True)
+    accllr_cond2, selection_time_cond2 = calc_activity_onset_accLLR(test_data_cond2, cond1, cond2, modality, bin_width, thresh_proportion=accllr_thresh, trial_average=True)
```
rcanfield2, line 1030: Not sure what to input here as 'cond1' and 'cond2' are

aopy/analysis.py
```diff
@@ -856,3 +854,179 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+    train_trial_mask = np.zeros(nbuild_trials, dtype=bool)
+    train_trial_mask[train_trialidxs] = True
+
+    train_data_cond1 = build_data_cond1[train_trial_mask,:]
+    train_data_cond2 = build_data_cond2[train_trial_mask,:]
+    valid_data_cond1 = build_data_cond1[np.logical_not(train_trial_mask),:]
+    valid_data_cond2 = build_data_cond2[np.logical_not(train_trial_mask),:]
+
+    # Input train and validate data to calculate acclllr threshold (transpose data back to input into accllr functions)
+    accllr_thresh = calc_accLLR_threshold(train_data_cond1.T, train_data_cond2.T, valid_data_cond2.T, modality, bin_width, step_size=step_size, false_alarm_prob=false_alarm_prob)
```
rcanfield2, line 1026: Should I be inputting (npt, ntrial) training data or trial average here to determine the training data to input

aopy/analysis.py
rcanfield2: Also, do you think splitting condition1 test and condition2 test datasets into different function calls is the right approach? I think it makes sense to me.

leoscholl: i don't know why we need to run accllr on condition1 data at all, but if you think it's useful then yes i agree this is the best approach.

aopy/analysis.py
leoscholl: should just be a single trial. i hadn't considered trial averaging but that would make sense, is that what they do in the paper?

rcanfield2:
A couple things to discuss when we meet:

1. We should change all cond1 and cond2 variable names to something including "null" and "alternative" hypothesis
2. We need to add signal quality compensation
3. Do we need to add the spike smoothing they did in the paper? (smooth each spike with a var=5ms Gaussian) We need to do something because if condition2 (null) has no activity for a time point the output accllr is NaN
4. Should we make this collection of functions a class?

tests/test_analysis.py
```diff
@@ -362,6 +365,96 @@ def test_linear_fit_analysis(self):
...
+        samplerate = 1
+        best_tp, tp, fa = aopy.analysis.calc_accLLR_threshold(altcond_train, nullcond_train, nullcond_test, modality='lfp', bin_width=1./samplerate, thresh_step_size=0.01, false_alarm_prob=0.05)
+        plt.plot(tp, fa)
+        plt.xlabel('Thresh proportion')
+        plt.ylabel('False Alarm Prop')
+        filename = 'accllr_thresh_prop.png'
+        aopy.visualization.savefig(write_dir, filename)
+
+    def test_calc_accLLR_wrapper(self):
+        cond2 = np.arange(100)
```
rcanfield2, line 443: Test multi-channel

aopy/analysis.py
```diff
@@ -856,3 +854,282 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+        LL(t)=log \\frac{P[dN(t))|\lambda_1(t)]}{P[dN(t))|\lambda_2(t)]}
+    
+    Args:
+        data (npts, ntrial): Input data to compare to each condition
+        altcond (npts): neural activity series of interest (condition 1)
+        nullcond (npts): neural activity during a baseline period for comparison (condition 2)
+        modality (str): Either 'lfp' or 'spikes'. 'lfp' uses a Gaussian distribution assumption and 'spikes' uses a Poisson distribution assumption
+        bin_width (float): Bin width of input activity
+        thresh_proportion (float): Proportion of maximum AccLLR where the threshold is set to classify trials as condition 1 or condition 2. 
+        max_accLLR (float):
```
leoscholl, line 892: can you explain what this is for?

aopy/analysis.py
```diff
@@ -856,3 +854,282 @@ def linear_fit_analysis2D(xdata, ydata, weights=None, fit_intercept=True):
...
+    '''
+    nt = data_altcond.shape[0]
+    nch = data_altcond.shape[1]
+    ntrials = data_altcond.shape[2]
+
+    # Split into model building (train, validate) and testing data sets (outputs (ntrial, npt) datasets)
+    # test_data_altcond, build_data_altcond, test_data_nullcond, build_data_nullcond = model_selection.train_test_split(data_altcond.T, data_nullcond.T, test_size=train_prop_input)
+    ntrain_trials = int(np.ceil(ntrials*0.4))
+    nvalid_trials = int(np.ceil(ntrials*0.3))
+    ntest_trials = ntrials - ntrain_trials - nvalid_trials
```
leoscholl, line 1059: currently this is hard-coded. we can do something like:
```suggestion
    ntrain_trials = int(np.ceil(ntrials*train_prop_input*0.7))
    nvalid_trials = int(np.ceil(ntrials*train_prop_input*0.3))
    ntest_trials = ntrials - ntrain_trials - nvalid_trials
```
so at least we have flexibility over the amount of data being used for training vs testing

rcanfield2:
I don't know how to test signal selectivity matching or if i'm even doing the selectivity matching correctly.

leoscholl (approved):
Seems pretty good, there is just one argument that doesn't have a description

PR #250, opened 2022-02-24
improve spatial maps

leoscholl:
fixed a problem with calc_data_map that it wasn't returning the new positions, and a bug with the calculation of xy positions.
added some tests and an example that explain how to use calc_data_map to interpolate spatial data


leoscholl:
also added an option for transparency to plot_spatial_map

pavi-rajes (approved):
changes look good to me

PR #256, opened 2022-03-12
remove trials and state as default bmi3d data

leoscholl:


m-nolan (approved):
Style is consistent, maintains utility function use in a consistent way vis a vis hdf5 record access and data dict management.

Issue #257, opened 2022-03-14
Add timing channel pulse detection code


m-nolan:
Add function for getting pulse times and pulse duty cycles from digital timing channels.

This will be useful for video frame alignment.

PR #258, opened 2022-03-14
Digital signal pulse width

m-nolan:
#257 Add code fulfilling issue 157 requirements


aopy/utils.py
```diff
@@ -300,6 +300,41 @@ def get_edges_from_onsets(onsets, pulse_width):
         values[2+2*t] = 0
     return timestamps, values
 
+def get_pulse_times( digital_data, sync_channel, trigger_channel, samplerate ):
```
leoscholl, line 303:
```suggestion
def get_e3v_frame_timestamps( digital_data, sync_channel, trigger_channel, samplerate ):
```

leoscholl (commented):
i think since this is now specific to e3v cameras (sync and trigger channels, 6 pulses, etc.), it is no longer a "utility" and more of a preprocessing function. i would rename it and maybe move it somewhere else. maybe separate the calculating pulse times and duty cycle into its own function?

m-nolan:
refactored the code as asked.

aopy/data.py
```diff
@@ -1,7 +1,11 @@
 # data.py
 # Code for directly loading and saving data (and results)
 
+from ctypes import util
+from random import sample
 import numpy as np
+
+from aopy import utils
```
leoscholl, line 8: this doesn't look right. should be relative import `from . import utils`

aopy/data.py
```diff
@@ -1069,3 +1074,34 @@ def yaml_read(filename):
...
+
+
+def get_e3v_video_frame_data( digital_data, sync_channel_idx, trigger_channel_idx, samplerate ):
+    
+    """get_e3v_video_frame_data
+
+    Compute pulse times and duty cycles from e3vision video data frames collected on an ecube digital panel.
+
+    Args:
+        digital_data (np.array): [n_time x n_channel] array of data read from ecube digital panel
```
leoscholl, line 1086: i prefer 
```suggestion
        digital_data (nt, nch): array of data read from ecube digital panel
```

aopy/utils.py
```diff
@@ -300,6 +300,44 @@ def get_edges_from_onsets(onsets, pulse_width):
         values[2+2*t] = 0
     return timestamps, values
 
+def get_pulse_edge_times( digital_data, samplerate ):
+
+    """get_pulse_edge_times
+
+    Args:
+        digital_data (np.array): [n_time x 1] array of data from ecube digital panel
```
leoscholl, line 308: same formatting as above

aopy/utils.py
```diff
@@ -300,6 +300,44 @@ def get_edges_from_onsets(onsets, pulse_width):
...
+
+def compute_pulse_duty_cycles( edge_pairs ):
+
+    """compute_pulse_duty_cycles
+
+    Args:
+        edge_pairs (np.array): [n_pulse x 2] start, end times from a series of pulses
+
+    Returns:
+        duty_cycle (): _description_
```
leoscholl, line 332: finish this docstring

aopy/data.py
```diff
@@ -1069,3 +1074,34 @@ def yaml_read(filename):
...
+
+    Returns:
+        pulse_times (np.array): array of floats indicating pulse start times
+        duty_cycle (np.array): array of floats indicating pulse duty cycle (quotient of pulse width and pulse period)
+    """
+
+    trig_pulse_edges = get_pulse_edge_times(digital_data[:,trigger_channel_idx],samplerate)
+    trig_pulse_times = trig_pulse_edges[:,0]
+    start_trig_time = trig_pulse_times[0]
+    end_trig_time = trig_pulse_times[3]
```
leoscholl, line 1099: a comment here about what the trigger signal looks like might be useful for people looking at this in the future

leoscholl (approved):
cool, just a few minor comments. also technically compute_pulse_duty_cycles( ) and get_pulse_edge_times() should have tests of their own. maybe write something simple. or maybe write an empty test and point people to the test for get_e3v_video_frame_data()

aopy/data.py
m-nolan: the format is `name (type): description`, right?

leoscholl: technically it's `varIn_name1 (dim label1, dim label2): Description`
see https://analyze.readthedocs.io/en/latest/documentation-guidelines.html

PR #263, opened 2022-03-21
fix plot_tuning_curves

rcanfield2:


m-nolan (approved):
Looks good. Make sure there are tests for these

PR #265, opened 2022-03-22
converted matlab functions related to mtfilter to python

resatomo:
converted 4 functions (dpsschk, dp_proj, mtfilt, mtfilter)


resatomo:
I tried filtering data with a narrow band, but it didn't work well in the case of a too narrow band. I added some examples, but I didn't know which examples should be included and kept in the test file.

leoscholl (approved):
seems good to me!

PR #268, opened 2022-03-23
Leo update calc eyetracking

leoscholl:
use interpolated cursor coordinates instead of relying on cycle counts


m-nolan (approved):

leoscholl:
thanks. i agree nothing is broken further, but still unsure whether things might be broken without us being aware... for context, eyetracking stopped working reliably a while ago and i narrowed the problems down to alignment between the eye tracking data and the bmi3d trial events. the main change here is the way the clock timestamps are preprocessed, now there is extra code to fill in missing data and i fixed a bug where the approximate timestamps were offset by some constant.

PR #278, opened 2022-04-08
Adding move_files() in data.py.

pavi-rajes:
This function is a part of backing up data in our server.


aopy/data.py
```diff
@@ -16,6 +16,34 @@
...
+            # print(dest)
+            if not os.path.exists(dest):
+                print(f"Copying {dest}...", end="")
+                if os.path.isfile(source):
+                    shutil.copy2(source, dest)
+                else:
+                    shutil.copytree(source,dest)
+                print("")
+            else:
+                print(f"File {source} already moved!")
```
leoscholl, line 44: i think files_list should be just a list, not a list of lists.
```suggestion
    for source in files_list:
        dest = os.path.join(dest_dir, os.path.basename(source))
        # print(dest)
        if not os.path.exists(dest):
            print(f"Copying {dest}...", end="")
            if os.path.isfile(source):
                shutil.copy2(source, dest)
            else:
                shutil.copytree(source,dest)
            print("")
        else:
            print(f"File {source} already moved!")
```

tests/test_data.py
```diff
@@ -460,5 +460,21 @@ def test_get_pulse_times(self):
...
 
+class Backup(unittest.TestCase):
+
+    def test_move_files(self):
+        source_dir = [['tests/data/test_source/dir_1'],['tests/data/test_source/dir_2']]
+        #source_files = [os.path.join(source_dir,'load_hdf_contents_test.hdf'), os.path.join(source_dir,'load_hdf_group_test.hdf')]
+
+        dest_dir = 'tests/tmp/test_destination'
+        if not os.path.exists(dest_dir):
+            os.mkdir(dest_dir)
```
leoscholl, line 471: you should also delete the destination file if it already exists

tests/test_data.py
```diff
@@ -460,5 +460,21 @@ def test_get_pulse_times(self):
...
+
+        move_files(source_dir,dest_dir)
+        dest_dir_file_bool1 = os.path.exists('tests/tmp/test_destination/dir_1')
+        dest_dir_file_bool2 = os.path.exists('tests/tmp/test_destination/dir_2')
+        self.assertTrue(dest_dir_file_bool1)
+        self.assertTrue(dest_dir_file_bool2)
```
leoscholl, line 477: also check if the files themselves have been copied, e.g. `self.assertTrue(os.path.exists('tests/tmp/test_destination/dir_1/load_ecube_data.png'))`

leoscholl (changes requested):
a few changes to make, then we can merge

tests/test_data.py
pavi-rajes: do you mean for the unit tests?

pavi-rajes: I have added code to empty our test destination prior to running unittest.

aopy/data.py
```diff
@@ -16,6 +16,35 @@
...
+            # print(dest)
+            if not os.path.exists(dest):
+                print(f"Copying {dest}...", end="")
+                if os.path.isfile(source):
+                    shutil.copy2(source, dest)
+                else:
+                    shutil.copytree(source,dest, dirs_exist_ok= overwrite)
+                print("")
+            else:
+                print(f"File {source} already moved!")
```
leoscholl, line 45: i tried using this to backup the files from our optogenetics experiments, i had a couple problems
- the destination directory needs to be the full path, including the system name. if you're copying arbitrary systems, this can be annoying
- if the system folder doesn't exist, it should be created

I came up with the following:

```
def move_files(files_list, source_dir, dest_dir):
    for file in files_list:
        source = os.path.join(source_dir, file)
        dest = os.path.join(dest_dir, file)
        dirname = os.path.dirname(dest)
        if not os.path.exists(dirname):
            os.mkdir(dirname)
        if not os.path.exists(dest):
            print(f"Copying {dest}...", end="")
            if os.path.isfile(source):
                shutil.copy2(source, dest)
            else:
                shutil.copytree(source,dest)
            print("")
        else:
            print(f"File {source} already moved!")
```

PR #279, opened 2022-04-08
Updating code for participation ratio

pavi-rajes:
Please check my code for estimating participation ratio. My current implementation lets the user chose between Eigen values or trace(cov) to estimate PR. 

Also, need to check if the equation shows up right in Sphinx documentation.

### DIFF
--- a/aopy/analysis.py
+++ b/aopy/analysis.py
@@ -17,7 +17,8 @@ from scipy import interpolate
 from scipy.optimize import curve_fit
 from scipy import stats, signal
 import warnings
-from numpy.linalg import inv as inv # used in Kalman Filter
+from numpy.linalg import inv as inv  # used in Kalman Filter
+import copy
 
 import warnings
 from . import preproc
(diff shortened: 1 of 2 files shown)


rcanfield2:
For some reason it won't let me comment on the code itself but just a couple things.

1. In the documentation I think we need to define which dimensions of the input matrix are observations or features. As written I think its (nfeat x nobs), which is transformed from what we normally to.
2. In the documentation it says the output PR is an 'int' but it can be a float
3. Does np.cov center the input data? I can't find anything in the documentation?

aopy/analysis.py
```diff
@@ -110,6 +110,32 @@ def get_pca_dimensions(data, max_dims=None, VAF=0.9, project_data=False):
 
     return list(explained_variance), num_dims, projected_data
 
+def estimate_PR(data, method = 'eig'):
```
rcanfield2, line 113: I wonder if the default should be calculated using the trace since I think it is more computationally efficient. If we use the 'eig' method we should also probably return the eigen vectors.

pavi-rajes: Hi Ryan, I made some code changes to the code. I removed 'trace' as a method in my analysis code just to stick to one method that works all the time. I ran into some complex numbers issue with 'trace' a couple of time. I know the current code works.

resatomo (approved):
Looks good.

PR #283, opened 2022-04-15
include examples

leoscholl:
add some test examples to the 'contributing' page


docs/source/test-guidelines.rst
```diff
@@ -75,3 +75,144 @@ And update the code to pass this test:
...
+    tapers = [N, NW]
+    x_mtfilter = precondition.mtfilter(x_312hz + x_600hz, tapers, fs=fs, f0=f0)
+    plt.figure()
+    plt.plot(x_312hz + x_600hz, label='Original signal (312 Hz + 600 Hz)')
+    plt.plot(x_312hz, label='Original signal (312 Hz)')
+    plt.plot(x_mtfilter, label='Multitaper-filtered signal')
+    plt.xlim([0,500])
+    plt.legend()
+    fname = 'mtfilter.png'
+    savefig(write_dir, fname) # Should have power in 312 Hz but not 600
```
rcanfield2, line 126: Is it possible to put this plot in the documentation?

docs/source/test-guidelines.rst
```diff
@@ -75,3 +75,144 @@ And update the code to pass this test:
...
+
+    result_filename = 'test_proc_eyetracking_short.hdf'
+    files = {}
+    files['ecube'] = '2021-09-30_BMI3D_te2952'
+    files['hdf'] = 'beig20210930_02_te2952.hdf'
+
+    # Should fail because no preprocessed experimental data
+    if os.path.exists(os.path.join(write_dir, result_filename)):
+        os.remove(os.path.join(write_dir, result_filename))
+    self.assertRaises(ValueError, lambda: proc_eyetracking(data_dir, files, write_dir, result_filename))
```
rcanfield2, line 140: What is going on here? If necessary some extra comments would be useful

rcanfield2 (approved):
Looks great. I made a couple comments on the code, mostly just adding more comments so the reader can easily understand what is going on. 

I think there should also be a couple sentences explaining what needs to be tested. All optional inputs, all outputs, etc.

Also, it might be useful to add some "theory" at the beginning. People may be avoiding test functions because they are overcomplicating them. Writing a couple sentences about the mental process you go through to determine what is an acceptable test could be useful. I try and distill the entire function to a couple types of results I need that would be sufficient to prove the test works. I then work backwards to create the minimal fake dataset.

PR #284, opened 2022-04-22
Add channel viz mike nolan

m-nolan:
code for visualizing individual channels of multichannel broadband/lfp/ecog data (trace, spectrogram, normalized spectrogram).


m-nolan:
Not sure why there are so many commits on that. I'll make a cleaner branch, so disregard this.

PR #285, opened 2022-04-22
Add: channel visualization

m-nolan:
multichannel data quality visualization code


aopy/visualization.py
```diff
@@ -934,13 +938,105 @@ def plot_boxplots(data, plt_xaxis, trendline=True, facecolor=[0.5, 0.5, 0.5], li
             whiskerprops=dict(color=linecolor), flierprops=dict(color=facecolor, markeredgecolor=facecolor),
             medianprops=dict(color=linecolor))
 
-def advance_plot_color(ax, n):
-    '''
-    Utility to skip colors for the given axis.
+
+def profile_data_channels(data, samplerate, figuredir):
+
```
leoscholl, line 943: docstrings needed for this and combine_channel_figures

leoscholl (changes requested):
some test functions would be nice, you can make a couple of example figures from the fake ecube data, perhaps. and then combine them into a 2x1 grid. for `profile_data_channels` you can just call it with the same data and make sure there aren't any errors. you can also include one of the test figures into /docs/source/images, and use it place of the ------------- diagram in `plot_channel_summary`

m-nolan:
addressed the following:
- replaced the `advance_plot_color` function I accidentally deleted
- added tests for the channel plots
- added an example to the docs source image directory ('channel_profile_example.png')

leoscholl (approved):
i added some suggestions -- mainly filling in more of the docstrings, also changed the clim in the example, but not sure why nothing shows up in the ch2 spectrogram. looks good overall

Issue #293, opened 2022-06-01
metadata load error in `.data.load_ecube_digital`


m-nolan:
https://github.com/aolabNeuro/analyze/blob/242d23bc36ee1db542567ddd8fa9a132b4ec9283/aopy/data.py#L405

This should be grabbing `'DigitalPanel'`, not `'AnalogPanel'`.

PR #294, opened 2022-06-01
Fix: metadata loading bug in .data.load_ecube_digital

m-nolan:
Addresses bug in issue #293


m-nolan:
Nevermind - wound up merging in material from a different branch accidentally. Ignore this.

PR #301, opened 2022-07-01
add plot_window_around_event

jasfouri:
work in progreso

### DIFF
--- a/aopy/analysis.py
+++ b/aopy/analysis.py
@@ -1156,3 +1156,59 @@ def interp_multichannel(x):
     x[nan_idx] = np.interp(idx,xp,fp)
 
     return x
+
+def plot_window_around_event(i_events, which_event, quality_name, quality_segments, quality, num_samples_before, num_samples_after, result):
+    '''
+    Plot a trace of cursor and eye directional error around a specified event, averaged across trials with a specified quality.
+    
+    Args:
+        i_events (ntrials): list of indices of event occurrence, for each trial
+        which_event (int): index of event in trial (e.g., 1 selects the second event in a trial)
+        quality_name (str): name of quality by which to select trials
+        quality_segments (ntrials): quality of each trial
+        quality (list): selected quality
+        num_samples_before (int): number of samples to plot before event
+        num_samples_after (int): number of samples to plot after event
+        result (str): desired result by which to select trials
+    
+    Returns:
+        None 
+    '''
+    segment_nums = [i for i, x in enumerate(quality_segments) if (x in quality) and (i in trial_result['reachsurround'][result])]
+    # inspect average directional error for all segments with one event where window is within bounds
+    segments_one_event = [i for i, x in enumerate(i_events) if i in segment_nums and len(x) > 0 and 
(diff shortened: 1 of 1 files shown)


aopy/postproc.py
```diff
@@ -511,4 +519,146 @@ def get_target_locations(preproc_dir, subject, te_id, date, target_indices):
...
+    Get the average x,y position data between left and right eyes
+    
+    Args:
+        eye_pos (nt, 4): array of eye position data for one trial, as output by exp_data
+    
+    Returns:
+        (nt, 2): array of x,y position data averaged between both eyes
+    '''
+    
+    return np.stack([(eye_pos[:,0]+eye_pos[:,2])/2, (eye_pos[:,1]+eye_pos[:,3])/2], axis=1)
```
leoscholl, line 586: could we incorporate the eye_labels into the input of this function so it knows exactly what indices to average?

aopy/postproc.py
```diff
@@ -511,4 +519,146 @@ def get_target_locations(preproc_dir, subject, te_id, date, target_indices):
...
+
+def plot_hist_and_get_threshold(data, num_sd = 0, xlabel=''):
```
leoscholl, line 588: it's slightly weird to have a function that computes something and also plots something. maybe you can make an optional argument to plot the histogram, but by default this function only computes the threshold?

aopy/postproc.py
```diff
@@ -511,4 +519,146 @@ def get_target_locations(preproc_dir, subject, te_id, date, target_indices):
...
+    relative_timestamps = old_event_timestamps - old_event_timestamps[0]
+    segment_duration = relative_timestamps[-1]  # in seconds
+    time_new = np.arange(np.ceil(segment_duration*new_samplerate))/new_samplerate
+    new_event_timestamps = [np.argmin(np.abs(time_new - x)) for x in relative_timestamps]  # in samples
+    if in_samples:
+        return new_event_timestamps
+    else:
+        return new_event_timestamps / new_samplerate  # in seconds
+
+def assign_zone(pos, target, target_radius):
```
leoscholl, line 635:
```suggestion
def assign_zone(pos, target, target_radius):
```
```suggestion
def assign_ja_zone(pos, target, target_radius):
```

aopy/analysis.py
```diff
@@ -510,7 +512,303 @@ def get_unit_spiking_mean_variance(spiking_data):
...
+            win_start[p] = np.max([0, win_start[p]])  # shrink windows to within bounds
+            win_end[p] = np.min([len(dist)-1, win_end[p]])
+            
+            if win_start[p] != win_end[p]:
+                dist_slope[p].append((dist[win_end[p]]-dist[win_start[p]])/(win_end[p]-win_start[p]))
+                vel_slope[p].append((vel[win_end[p]]-vel[win_start[p]])/(win_end[p]-win_start[p]))
+                
+    return cursor_dists, cursor_vels, dist_slope, dist_avg, vel_slope, vel_avg
+
+def compute_directional_error(pos, target, use_initial_direction = False):
```
leoscholl, line 688: There’s an existing function get_inst_target_dir which I think does the same thing as this. Maybe take a look before writing any tests for this one.

PR #302, opened 2022-07-08
Update postproc.py

pavi-rajes:
few minor updates only to documentation.


leoscholl:
Thanks, looks good except datatype is still optional and defaults to ‘cursor’

PR #305, opened 2022-07-13
add plot_ecog_map

leoscholl:
plotting function that is specific to ECoG244 data


aopy/visualization.py
```diff
@@ -277,6 +278,42 @@ def plot_spatial_map(data_map, x, y, alpha_map=None, ax=None, cmap='bwr'):
...
+    '''
+    
+    if ax is None:
+        ax = plt.gca()
+    
+    # Load the signal path files
+    elec_pos, acq_ch, elecs = load_chmap(drive_type='ECoG244')
+
+    # Remove bad electrodes
+    bad_ch = acq_ch[np.isin(elecs, bad_elec)]-1
```
jasfouri, line 303: I've been declaring acq_ch_idx and bad_ch_idx because I kept forgetting to subtract the 1 when indexing. Might be good to do here!

PR #311, opened 2022-07-20
Behavior metrics

pavi-rajes:
I added two functions:  time_to_target() and path_length() in analysis.py 
time_to_target() calculates reach time provided event codes and event times with an option to get per target statistics. Path length calculates the total distance traveled from cursor x and cursor y positions changing across time.  

In the same pull request, also adding another function get_trail_segments_and_times() in preproc.base. This function returns all corresponding event times along with event codes. 

Notes: 
Currently, time_to_target() only estimates reach times on successful targets. In the future, we can make it optional by adding an optional lamba reward filter function in the input. But I will treat this as an enhancement and this pull request need not wait for it.

### DIFF
--- a/aopy/analysis.py
+++ b/aopy/analysis.py
@@ -1582,12 +1582,12 @@ def time_to_target(event_codes, event_times, target_codes=list(range(81, 89)) , 
             | **reachtime_pertarget (list):** mean reach time per target
             | **trial_id (list):** target index on each trial
     '''
-    event_times = np.array([event_times[iTr] for iTr in range(len(event_times)) if reward_code in event_codes[iTr]])
-    event_codes = np.array([event_codes[iTr] for iTr in range(len(event_times)) if reward_code in event_codes[iTr]])
-    leave_center_idx = np.argwhere(event_codes == go_cue_code)[0, 1]
-    reach_target_idx = np.argwhere(np.isin(event_codes[0], target_codes))[0][0] # using just the first trial to get reach_target_idx
-    reachtime = event_times[:, reach_target_idx] - event_times[:, leave_center_idx]
-    target_dir = event_codes[:,reach_target_idx]
+    tr_T = np.array([event_times[iTr] for iTr in range(len(event_times)) if reward_code in event_codes[iTr]])
+    tr_E = np.array([event_codes[iTr] for iTr in range(len(event_times)) if reward_code in event_codes[iTr]])
+    leave_center_idx = np.argwhere(tr_E == go_cue_code)[0, 1]
+    reach_target_idx = np.argwhere(np.isin(tr_E[0], target_codes))[0][0] # using just the first trial to get reach_target_idx
+    reachtime = tr_T[:, reach_target_idx] - tr_T[:, leave_center_idx]
+    target_dir = tr_E[:,reach_target_idx]
 
     return reachtime, target_dir
 
--- a/aopy/analysis.py
+++ b/aopy/analysis.py
@@ -1563,28 +1563,41 @@ def compute_path_length_per_trial(trajectory):
     path_length = np.sum(lengths)
     return path_length
 
-def time_to_target(event_codes, event_times, get_stats_per_target= False):
+
+def time_to_target(event_codes, event_times, per_target_stats=False):
     '''
-    This function calculates
+    This function calculates reach time to target only on rewarded trials given event codes and event times.
+
     Args:
-        event_codes (list) :
-        event_times (list):
-        target_direction (list):
-        get_stats_per_target (bool):
+        event_codes (list) : event codes
+        event_times (list) : event times corresponding to the event codes. These event codes and event times could be the output of preproc.base.get_trial_segments_and_times().
+         get_stats_per_target (bool): optional, use if you want to calculate reach time per target
 
     Returns:
-
+        reach_times (list):
+        reach_times_per_target(list of lists) : optional if per_target_stats == True
     '''
     CURSOR_ENTER_PERIPHERAL_TARGET = list(range(81, 89))
     CENTER_TARGET_OFF = 32
     REWARD = 48
     tr_e = np.array([event_codes[iTr] for iTr in range(len(event_times)) if REWARD in event_codes[iTr]])
-    tr_t = np.array([event_times[iTr] for iTr in range(len(event_times)) if  REWARD in event_codes[iTr]])
-    leave_center_idx = np.argwhere(tr_e == CENTER_TARGET_OFF)[0]
-    reach_target_idx = np.argwhere(tr_e == CURSOR_ENTER_PERIPHERAL_TARGET)[0]
+    tr_t = np.array([event_times[iTr] for iTr in range(len(event_times)) if REWARD in event_codes[iTr]])
+    leave_center_idx = np.argwhere(tr_e == CENTER_TARGET_OFF)[0, 1]
+
+    reach_target_idx = np.argwhere(np.isin(tr_e[0], CURSOR_ENTER_PERIPHERAL_TARGET))[0][0]
+
     rt = tr_t[:, reach_target_idx] - tr_t[:, leave_center_idx]
 
     # mean reach time per target
-    target_dir = tr_e[reach_target_idx] - 80
-    for iT in np.unique(target_dir):
+    if per_target_stats:
+        rt_pertarget = []
+        trial_id = []
+        target_dir = tr_e[:, reach_target_idx]
 
+        for iT in np.unique(target_dir):
+            dir_idx = np.where(target_dir == iT)[0]
+            rt_pertarget.append(rt[dir_idx])
+            trial_id.append(dir_idx)
+        return rt, (rt_pertarget, trial_id)
+    else:
+        return rt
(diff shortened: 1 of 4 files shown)


aopy/analysis.py
```diff
@@ -1543,3 +1543,61 @@ def interp_multichannel(x):
     x[nan_idx] = np.interp(idx,xp,fp)
 
     return x
+
+
+'''
+Behavioral metrics 
+'''
+
+def compute_path_length_per_trial(trajectory):
```
leoscholl, line 1552: why is it per trial? doesn't this just take a single trajectory

aopy/analysis.py
```diff
@@ -1543,3 +1543,61 @@ def interp_multichannel(x):
...
+
+
+def time_to_target(event_codes, event_times, per_target_stats=False):
+    '''
+    This function calculates reach time to target only on rewarded trials given event codes and event times.
+
+    Args:
+        event_codes (list) : event codes
+        event_times (list) : event times corresponding to the event codes. These event codes and event times could be the output of preproc.base.get_trial_segments_and_times().
+         get_stats_per_target (bool): optional, use if you want to calculate reach time per target
```
leoscholl, line 1574: docstring arg doesn't match the input variable

aopy/analysis.py
```diff
@@ -1543,3 +1543,61 @@ def interp_multichannel(x):
...
+
+    Returns:
+        reach_times (list):
+        reach_times_per_target(list of lists) : optional if per_target_stats == True
+    '''
+    CURSOR_ENTER_PERIPHERAL_TARGET = list(range(81, 89))
+    CENTER_TARGET_OFF = 32
+    REWARD = 48
```
leoscholl, line 1582: these should be inputs to the function in case they change for other experiments

aopy/analysis.py
```diff
@@ -1543,3 +1543,61 @@ def interp_multichannel(x):
...
+    tr_e = np.array([event_codes[iTr] for iTr in range(len(event_times)) if REWARD in event_codes[iTr]])
+    tr_t = np.array([event_times[iTr] for iTr in range(len(event_times)) if REWARD in event_codes[iTr]])
```
leoscholl, line 1584: why is it necessary to restrict to rewarded trials? maybe leave this up to the user?

aopy/analysis.py
```diff
@@ -1543,3 +1543,61 @@ def interp_multichannel(x):
...
+    return path_length
+
+
+def time_to_target(event_codes, event_times, per_target_stats=False):
+    '''
+    This function calculates reach time to target only on rewarded trials given event codes and event times.
+
+    Args:
+        event_codes (list) : event codes
+        event_times (list) : event times corresponding to the event codes. These event codes and event times could be the output of preproc.base.get_trial_segments_and_times().
```
leoscholl, line 1573: are these lists of events or lists of lists of events, where each list is a trial? in other words, is the input to this function supposed to be trial separated already?

aopy/analysis.py
```diff
@@ -1543,3 +1543,61 @@ def interp_multichannel(x):
...
+    CURSOR_ENTER_PERIPHERAL_TARGET = list(range(81, 89))
+    CENTER_TARGET_OFF = 32
+    REWARD = 48
+    tr_e = np.array([event_codes[iTr] for iTr in range(len(event_times)) if REWARD in event_codes[iTr]])
+    tr_t = np.array([event_times[iTr] for iTr in range(len(event_times)) if REWARD in event_codes[iTr]])
+    leave_center_idx = np.argwhere(tr_e == CENTER_TARGET_OFF)[0, 1]
+
+    reach_target_idx = np.argwhere(np.isin(tr_e[0], CURSOR_ENTER_PERIPHERAL_TARGET))[0][0]
+
+    rt = tr_t[:, reach_target_idx] - tr_t[:, leave_center_idx]
```
leoscholl, line 1589: could you use slightly longer variable names -- all the rt / tr_t / tr_e is making it a bit hard to read

leoscholl (changes requested):
some comments to address but overall looks like it works!

pavi-rajes:
> some comments to address but overall looks like it works!

:) surely it works!! 

As I mentioned in the initial pull request, I will treat generalizing the reach time to include any trials as an enhancement to work in a separate pull request, since the function will break if the trial aligned event codes matrix is not of same length. We could include some checks or a condition that it should be a n_trials * 5 matrix, for example. I will fix your other comments on variable names. Thanks for reviewing!

aopy/analysis.py
pavi-rajes: Each trajectory is a cursor trajectory per trial. This function calculates path length per trial instead of path length for all trials in a session. Hence, the name says _per_trial

aopy/analysis.py
pavi-rajes: fixed.

aopy/analysis.py
pavi-rajes: fixed in a new commit

aopy/analysis.py
pavi-rajes: fixed

aopy/analysis.py
pavi-rajes: Yes, input to this function is trail-aligned event codes. Since we have another function to trial align event codes and time. It could be done outside the function.

pavi-rajes:
Hold on on review. Some tests failed. I will fix it tomorrow.

aopy/analysis.py
leoscholl: i would call this one "calc_path_length" and the one that does all trials in a session "calc_path_length_per_trial". to me it makes more sense to have the "_per_trial" suffix on a function that cares about trials. this one you just give it a path and it computes the length, it doesn't need to know anything about trial structure.

aopy/analysis.py
leoscholl: ok, can you please make it more clear in the docstring what the inputs are supposed to be? or give an example in the docstring?

aopy/analysis.py
pavi-rajes: I think _per_trial makes it explicit, and the function explanation says it is for a single trajectory. Other function names like time_to_target() or get_trial_segments() without _per_trial suffix are all used for batch processing. If that makes sense. I want to leave it as it is. If this causes further confusion, we can change it in a future patch.

aopy/analysis.py
pavi-rajes: Already updated function documentation in a commit yesterday to make it clear. Please check latest version and see if further explanation is required.

aopy/analysis.py
pavi-rajes: It could be '_per_trajectory' insted of 'per_trial' perhaps. But either way i think a suffix if important.

leoscholl (approved):

aopy/analysis.py
pavi-rajes: haha. I will update the code to reflect this.

aopy/analysis.py
```diff
@@ -1543,3 +1543,63 @@ def interp_multichannel(x):
...
+        only_rewarded_trials (bool, optional): only include trials with a reward code
+        per_target_stats (bool, optional): use if you want to calculate reach time per target
+
+    Returns:
+        reach_times (list):
+        reach_times_per_target(list of lists) : optional if per_target_stats == True
+    '''
+    event_times = np.array(event_times)
+    event_codes = np.array(event_codes)
+    if only_rewarded_trials:
```
pavi-rajes, line 1586: I am not sure I understand why this is helpful. The rest of the calculation outside the if loop will break if event_times and event_codes are a list of lists that are not of the uniform shape.

pavi-rajes: One way of including a more generic trial filter is to get it as a lambda function instead.

pavi-rajes: To be more clear, the function breaks because of some unsuccessful trials, such as the ones with center hold penalty with have shorter event_codes and event_times length, in which case, the rest of the calculations won't work. the condition for time to target is basically that the cursor enters the peripheral target.
This is what the lambda function could allow the user to choose. I hope my point gets across.

leoscholl: hmm i see what you mean.  is it always going to be the case that rewarded trials are the same length? maybe we should just compute the lengths of the trials one at a time to avoid this issue in all cases?

pavi-rajes: yes, always rewarded trials will have same event code sequence and will be of same length. I think we can keep the fucntion as it was before and add a note saying this will only work for rewarded trials. 
I will soon be generalizing to unrewarded trials with full trajectories as well. Like I said before, I will treat that as an enhancement.

leoscholl: i added another function called `calc_segment_duration` which does similar calculations but on any segment. i also found a bug in your code if there are more than 1 of the same target then the per target stats are wrong, they only take the first appearance of each target into account. i think you need to take a mean somewhere.

pavi-rajes: Could you explain a use case or the need for another function here? The time_to_target() does what it says it does. It would not be possible for me to test edge case scenarios extensively. However, I know this function works with our beignet's data and is generic enough to be applied to any session or task. 
The event codes for cursor entering peripheral targets are supposed to be unique. This function works as long as the codes are unique. Taking means within function would change the statistics when we concatenate reach times for example, so I avoided that.

PR #312, opened 2022-08-01
add trial average option

leoscholl:
hi, adding an option to the calc_max_erp() function to return a max erp for every trial


resatomo (approved):
Looks good.

PR #317, opened 2022-09-09
Fix hand traj outliers

rcanfield2:


rcanfield2:
I have no idea where to put this in the preprocessing pipeline

leoscholl:
> I have no idea where to put this in the preprocessing pipeline

Line 400 of preproc/bmi3d.py is where the interpolated cursor data gets added, that's probably where i would include this as well.

rcanfield2:
@leoscholl Can you check that I added the filtering to the right place in the preprocessing

aopy/preproc/bmi3d.py
```diff
@@ -408,6 +408,8 @@ def _prepare_bmi3d_v1(data, metadata):
         data['cursor_interp'] = cursor_data_time
         metadata['cursor_interp_samplerate'] = cursor_samplerate
```
leoscholl, line 410:
```suggestion

    # In some versions of BMI3D, hand position contained erroneous data
    # caused by `np.empty()` instead of `np.nan`. The 'clean_hand_position' 
    # replaces these bad data with `np.nan`.
```

leoscholl (approved):

PR #321, opened 2022-09-22
update spatial map alpha

### LINKED ISSUE #282: plot_spatial_map alpha_map does not work
matplotlib 3.3.1 no longer supports using an array of alpha values, instead it wants RGBA values in the image. needs to be updated or revert to matplotlib 3.1.3

leoscholl:
to close #282 i've updated plot_spatial_map to convert the data into rgba and apply the alpha layer there.


aopy/visualization.py
```diff
@@ -247,6 +248,22 @@ def plot_spatial_map(data_map, x, y, alpha_map=None, ax=None, cmap='bwr'):
```
mmadduri, line 248: could you include what clim and show_bad are?

aopy/visualization.py
```diff
@@ -261,18 +278,24 @@ def plot_spatial_map(data_map, x, y, alpha_map=None, ax=None, cmap='bwr'):
...
+        norm = cm.colors.Normalize(*clim)
+        scalarMap = cm.ScalarMappable(norm=norm, cmap=cmap)
+        data_map = scalarMap.to_rgba(data_map)
+
+        # Apply the alpha map after scaling from 0 to 1
         alpha_range = np.nanmax(alpha_map) - np.nanmin(alpha_map)
         alpha_map = (alpha_map - np.nanmin(alpha_map)) / alpha_range
         alpha_map[np.isnan(alpha_map)] = 0
-
+        data_map[:,:,3] = alpha_map
```
mmadduri, line 293: is it always 3? would it be more robust to add a new axis or something?

mmadduri:
@leoscholl added some minor comments and will approve after you respond!

aopy/visualization.py
leoscholl: yeah it's always 3, rgb are 0-2 and 3 is alpha. i think this is ok

leoscholl:
thanks for the comments. good catch on the docstring! one of those parameters didn't do anything.. should be fixed now

mmadduri (approved):
looks good, thanks!

PR #323, opened 2022-10-05
add calc mean across adjacent electrodes

resatomo:
I added code computing the mean across adjacent electrodes


leoscholl (approved):
It looks fine

leoscholl:
could you separate this into a get_adjacent_electrodes() function, and then taking the mean? It would be useful to know exactly which electrodes are being averaged

PR #326, opened 2022-10-10
Leo update filter tests

resatomo:
modified code


leoscholl (approved):
Thanks for the help. I changed a lot of things, can you double check that it still looks ok compared to the matlab version?

resatomo:
I noticed I didn't add conj() in only 1 spot, so I added it. But this conj() didn't seem to be important. I also compared the code with Matlab version and it looks okay.

PR #329, opened 2022-10-12
correct both onset and offset times

leoscholl:
Adjusted slightly the algorithm to check both onset and offset times independently first, then combine them later in case there are different number of ons and offs

### DIFF
--- a/aopy/preproc/bmi3d.py
+++ b/aopy/preproc/bmi3d.py
@@ -567,7 +567,7 @@ def get_laser_trial_times(preproc_dir, subject, te_id, date, debug=False, **kwar
         trials = exp_data['bmi3d_trials']
         powers = trials['power'][:len(times)]
         edges = trials['edges'][:len(times)]
-        widths = [t[1] - t[0] for t in edges]
+        widths = np.array([t[1] - t[0] for t in edges])
         thr_width=0.001
         thr_power=0.05
     else:
--- a/aopy/preproc/bmi3d.py
+++ b/aopy/preproc/bmi3d.py
@@ -465,7 +465,7 @@ def find_laser_stim_times(laser_event_times, laser_event_widths, laser_event_pow
             | **corrected_times (nevent):** corrected laser timings (seconds)
             | **corrected_widths (nevent):** corrected laser widths (seconds)
             | **corrected_powers (nevent):** corrected laser powers (fraction of maximum)
-            | **times_not_found (nevent):** boolean array of times where no sensor measurement was found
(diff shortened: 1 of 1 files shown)


aopy/preproc/bmi3d.py
```diff
@@ -465,7 +465,7 @@ def find_laser_stim_times(laser_event_times, laser_event_widths, laser_event_pow
             | **corrected_times (nevent):** corrected laser timings (seconds)
             | **corrected_widths (nevent):** corrected laser widths (seconds)
             | **corrected_powers (nevent):** corrected laser powers (fraction of maximum)
-            | **times_not_found (nevent):** boolean array of times where no sensor measurement was found
+            | **times_not_found (nevent):** boolean array of times where no sensor measurement was found (either onset or offset)
```
rcanfield2, line 468: From reading this description it sounds like onset and offsets are different times but from the code it looks like this is a boolean array of laser pulses that are missing either an onset or offset sensor measurement.

aopy/preproc/bmi3d.py
```diff
@@ -481,33 +481,34 @@ def find_laser_stim_times(laser_event_times, laser_event_widths, laser_event_pow
...
-    adjusted_powers[~np.isnan(corrected_idx)] = laser_sensor_powers[valid_idx]
+    # Normalize sensor power to the highest power, then multiply by the highest trial power
+    laser_sensor_powers /= np.max(laser_sensor_powers)
+    laser_sensor_powers *= np.max(laser_event_powers)
 
     # Correct the widths and powers with the given thresholds
-    corrected_widths, widths_above_thr = validate_measurements(laser_event_widths, adjusted_widths, thr_width)
-    corrected_powers, powers_above_thr = validate_measurements(laser_event_powers, adjusted_powers, thr_power)
+    corrected_widths, widths_above_thr = validate_measurements(laser_event_widths, laser_sensor_widths, thr_width)
+    corrected_powers, powers_above_thr = validate_measurements(laser_event_powers, laser_sensor_powers, thr_power)
```
rcanfield2, line 511: If i'm understanding the code correctly this assumes that np.max(laser_event_powers) is the same as np.max(laser_sensor_powers). Is this okay?

rcanfield2 (approved):

aopy/preproc/bmi3d.py
leoscholl: it is supposed to *force* them to be the same. i figure the sensor measurement is meaningless anyway, so just scale it to match the given event power

PR #337, opened 2022-11-01
add memory management utils

leoscholl:


aopy/utils.py
```diff
@@ -429,4 +431,54 @@ def calc_radial_dist(pos, origin=(0,0)):
...
+    resource.setrlimit(resource.RLIMIT_AS, (maxsize, -1))
+
+def get_memory_limit_gb():
+    '''
+    Get the memory resource limit in gigabytes.
+
+    Returns:
+        int or None: upper limit of memory available to python in gigabytes
+    '''
+    soft, hard = resource.getrlimit(resource.RLIMIT_AS)
```
pavi-rajes, line 475: Adding some more information about the soft and hard limits. The soft limit is a value upon which the operating system will notify the process via a signal or typically restrict it. An upper bound on the memory values is defined by the hard limit. Although the hard limit can be lowered, it can never be raised by user processes (even if the process lowered itself) and is controlled by a system-wide parameter set by the system administrator.

aopy/utils.py
```diff
@@ -429,4 +431,54 @@ def calc_radial_dist(pos, origin=(0,0)):
...
\ No newline at end of file
+    return dist
+
+'''
+Memory
+'''
+def get_memory_available_gb():
+    '''
+    Get the available system memory in gigabytes. Only works on linux platforms.
+
```
pavi-rajes, line 442: Just added some more info: The results of this function are equivalent to this command in the terminal. `grep MemAvailable /proc/meminfo ` -> available memory 
`grep MemTotal /proc/meminfo`  -> total memory

pavi-rajes (approved):
The functions look good. I just added some additional information that might be useful to be included in the documentation. It is really just extra information but could be handy to use and easy to find in the documentation when one looks up memory functions.

PR #352, opened 2022-12-21
add assertions to proc_single

leoscholl:
so that we can better detect errors in pre-processing. i also added downsampling to the eyetracking data since it's wasting a lot of space at 25khz


leoscholl:
Changing the eye sampling rate highlights a problem with `get_kinematic_segements()`: how are we supposed to know the sampling rate of the data returned? does it matter?

aopy/preproc/oculomatic.py
```diff
@@ -1,6 +1,7 @@
+from .. import precondition
 from .. import data as aodata
 
-def parse_oculomatic(data_dir, files, debug=True):
+def parse_oculomatic(data_dir, files, samplerate=480, debug=True):
```
rcanfield2, line 4: Why choose an oculomatic sampling rate of 480? It might be useful to default to the sampling rate of other data

rcanfield2:
> Changing the eye sampling rate highlights a problem with `get_kinematic_segements()`: how are we supposed to know the sampling rate of the data returned? does it matter?

We should probably have it output by the function.

rcanfield2 (approved):
Looks good! Sorry I forgot about this.

PR #354, opened 2022-12-23
Leo update get target locations

leoscholl:
using `get_target_locations()` is tricky because if you try to get target 10 and there are only 9 targets, it gives you a cryptic error about index 0 being out of bounds. I added a more descriptive error message to help with debugging.


pavi-rajes:
useful update. Looking good.

pavi-rajes (approved):
thanks for the update. Error message is clear.

PR #355, opened 2023-01-13
Fix pause bug

leoscholl:
Hopefully. I'm checking the minimum time between pulses, if it's less than the pulse width then there must be problems and I run the new function to extend all the pulses to at least the pulse width.


katherineperks:
error in test_data (may be related to downsampling of eye data): 
image

error in test_preproc:
image

katherineperks (approved):
Looks good. I'm not sure that detect_edges can accurately find falling edges when given a min_pulse_width, but it looks like you set falling to False in preproc/bmi3d.py so it shouldn't cause problems.

PR #357, opened 2023-01-24
Fix no event bug

leoscholl:
There was a bug for experiments that don't have sync events that prevented preprocessing. I added a check of the length of sync events to avoid this problem.


katherineperks:
Does the file you're using to test this (id=6890) have no sync events in the raw data? When I load that entry's preprocessed data, it has many sync events.

katherineperks:
Looks good

katherineperks (approved):

PR #359, opened 2023-03-03
Leo eyetracking

leoscholl:
work in progress


aopy/precondition/eye.py
```diff
@@ -0,0 +1,117 @@
...
+    eye_pos = eye_pos[n_pad:-n_pad,:]
+    
+    # Differentiate twice to get acceleration
+    time = np.arange(eye_pos.shape[0])/samplerate
+    velocity = derivative(time, eye_pos, norm=True)
+    accel = derivative(time, velocity, norm=False)
+
+    # Set an appropritate threshold to detect saccades
+    baseline_mean = np.mean(accel)
+    baseline_std = np.std(accel)
```
resatomo, line 62: How about setting threshold as the argument? The current code computes the threshold at each trial, but if you want to compare the number of saccades across trials and days, I thought it's better to change this code so that we can set the same threshold across trials and days. I don't know how much changes in baseline there is across trials and days , but there might be big changes in baseline before and after learning in difficult tasks such as BMI control.

aopy/precondition/eye.py
```diff
@@ -0,0 +1,180 @@
...
+        return [], [], []
+
+    # Only consider saccades that have a threshold-crossing offset
+    onset = []
+    duration = []
+    distance = []
+    prev_offset = 0
+    for t in saccade_onset_time:
+        if intersaccade_min and t < prev_offset + intersaccade_min:
+            continue # ignore saccades within intersaccade_min of the last saccade
```
resatomo, line 140: intersaccade_min doesn't have any roles?

resatomo (approved):
I think it looks good. Thanks for writing all code.

aopy/precondition/eye.py
leoscholl: Yeah by default, saccades can be right next to each other. I found in some data that there are double saccades and if you ignore one of them you miss a big part of the data. But if you set intersaccade_min as 20ms for instance then it will only count the first one in the double.

PR #362, opened 2023-03-06
Leo copy edges 2

leoscholl:
Only make edges longer, not shorter, by or-ing the original data with the new data


katherineperks (approved):
Looks good!

PR #363, opened 2023-03-06
add function to calculate local correlation map with a moving square kernel.

rcanfield2:


aopy/analysis/base.py
```diff
@@ -867,3 +867,72 @@ def interp_multichannel(x):
...
+                                        boundary='fill', mode='same')
+        irow, icol = np.unravel_index(np.argmax(corr), corr.shape)  # find the match
+        row_shift = int(irow - (data1.shape[0]-1)/2)
+        col_shift = int(icol - (data1.shape[1]-1)/2)
+        data2_align = np.roll(data2, row_shift, axis=0)
+        data2_align = np.roll(data2_align, col_shift, axis=1)
+        shifts = (row_shift, col_shift)
+    else:
+        data2_align = data2
+        shifts = None
```
leoscholl, line 913: could shifts be (0,0) by default so it's always a valid output?

aopy/analysis/base.py
```diff
@@ -867,3 +867,72 @@ def interp_multichannel(x):
...
+            data_subset1 = data1_pad[(xx-start_idx):(xx+start_idx+1),(yy-start_idx):(yy+start_idx+1)]
+            data_subset2 = data2_pad[(xx-start_idx):(xx+start_idx+1),(yy-start_idx):(yy+start_idx+1)]
+
+            # If either data subset is all 0's set the NCC to 0
+            if np.linalg.norm(data_subset1)==0 or np.linalg.norm(data_subset2)==0:
+                NCC[xx - start_idx, yy - start_idx]
+            else:
+                data_subset1 /= np.linalg.norm(data_subset1)
+                data_subset2 /= np.linalg.norm(data_subset2)
+                NCC[xx - start_idx, yy - start_idx] = np.max(scipy.signal.correlate2d(data_subset1,data_subset2))
```
leoscholl, line 933: not sure i understand the np.max here, could you explain in comments?

leoscholl (commented):
I think an example would be really useful here, could you include the test map as a figure that shows up in the docs?

aopy/analysis/base.py
rcanfield2: scipy.signal.correlate2d computes the cross correlation between the two arrays input. I couldn't find a python function to just compute the 2d correlation value. We should probably be taking correlation at 0 lag though.

leoscholl (approved):
This looks good!

PR #366, opened 2023-03-10
add concat trials function

leoscholl:
function to concatenate trials (hand, eye, or cursor trajectories) across multiple recordings


aopy/data/bmi3d.py
```diff
@@ -564,4 +566,127 @@ def get_source_files(preproc_dir, subject, te_id, date):
...
+        if include_handdata:
+            hand_traj = get_kinematic_segments(preproc_dir, subject, te, date, trial_start_codes, trial_end_codes, 
+                                                  datatype='hand')[0].tolist()
+        eye_traj = [None] * len(tr_seg)
+        if include_eyedata:
+            eye_traj = get_kinematic_segments(preproc_dir, subject, te, date, trial_start_codes, trial_end_codes, 
+                                                 datatype='eye')[0].tolist()
+
+        target_idx = [code[np.isin(code, target_codes)][0] - target_codes[0] if np.sum(np.isin(code, target_codes)) == 1 else 0 for code in tr_seg]
+        target_location = get_target_locations(preproc_dir, subject, te, date, target_idx).tolist()
```
rcanfield2, line 629: What is the point of including this?

aopy/data/bmi3d.py
```diff
@@ -564,4 +566,127 @@ def get_source_files(preproc_dir, subject, te_id, date):
             |** data_dir (str):** directory where the source files were located
     '''
     exp_data, exp_metadata = load_preproc_exp_data(preproc_dir, subject, te_id, date)
-    return exp_metadata['source_files'], exp_metadata['source_dir']
\ No newline at end of file
+    return exp_metadata['source_files'], exp_metadata['source_dir']
+
+def concat_trials(preproc_dir, subjects, ids, dates, trial_start_codes, trial_end_codes, target_codes,
```
rcanfield2, line 571: I think either the title should specify that this is for task data or the function should be written to concatenate trials of any variable name in the preprocessed data (including neural data such as lfp_data).

aopy/data/bmi3d.py
```diff
@@ -564,4 +566,127 @@ def get_source_files(preproc_dir, subject, te_id, date):
...
+                                         'eye': eye_traj,
+                                        })], ignore_index=True)
+    
+    return df
+
+def concat_trials_center_out(preproc_dir, subjects, ids, dates, df=None, include_center_target=True,
+                  include_handdata=False, include_eyedata=False):
+    '''
+    Wrapper around concat_trials() specifically for center-out experiments.
+
```
rcanfield2, line 654: Can you include more of the assumptions of the function in the description. It seems like this is mostly just restricting the event codes used.

aopy/data/bmi3d.py
```diff
@@ -564,4 +566,127 @@ def get_source_files(preproc_dir, subject, te_id, date):
...
+                                         'date': date, 
+                                         'event_codes': tr_seg,
+                                         'event_times': tr_t, 
+                                         'reward': reward,
+                                         'penalty': penalty,
+                                         'target_idx': target_idx,
+                                         'target_location': target_location,
+                                         'cursor': cursor_traj, 
+                                         'hand': hand_traj, 
+                                         'eye': eye_traj,
```
rcanfield2, line 645: Is there a reason the dataframe name drops the 'traj'? Dropping 'traj' leaves room for ambiguity about what data it is (position vs. velocity)

rcanfield2: Speaking of, is it worth generalizing this function to allow the user to request either position or velocities?

aopy/data/bmi3d.py
```diff
@@ -564,4 +566,127 @@ def get_source_files(preproc_dir, subject, te_id, date):
...
+    Wrapper around concat_trials() specifically for center-out experiments.
+
+    Args:
+        preproc_dir (str): base directory where the files live
+        subjects (list of str): Subject name for each recording
+        ids (list of int): Block number of Task entry object for each recording
+        dates (list of str): Date for each recording
+        df (DataFrame, optional): pandas DataFrame object to append. Defaults to None.
+        include_center_target (bool, optional): If True, trials begin after the cursor enters
+            the center target. Otherwise trials begin after the go cue. Default True.
```
rcanfield2, line 662: Is there a way to also allow the user to pick when the cursor is leaves the center target?

rcanfield2 (commented):
Correct me if I'm wrong, but these functions don't actually concatenate the trial segments. It looks to me like they extract the trial segments and place them all in a dataframe with their respective information. I tend to think of concatenation as when many trials become one continuous array.

aopy/data/bmi3d.py
leoscholl: Because target idx can (and does) change across recordings, so if we don’t add it here it’s a pain to go through each row to add it later. But I do want this function to be useful for other kinds of trials, eg tracking. Not sure how this would work with those.

aopy/data/bmi3d.py
leoscholl: I did try to include lfp segments but it wasn’t very easy and probably doesn’t make much sense. I think this function should be limited to kinematics for now. It might be useful to separate out the trial concatenation from the kinematics part, so we could reuse the trial concatenation for other kinds of data. But maybe outside the scope of this pull request.

aopy/data/bmi3d.py
leoscholl: Will do

aopy/data/bmi3d.py
leoscholl: No reason, I will add them back. Yeah it would be great to have some preprocessing of the kinematics in case you want them in velocity eg. Might be a bit messy.

aopy/data/bmi3d.py
leoscholl: Leave center target is roughly equivalent to center target turning off. If you wanted cursor leaving the center target radius, you could start with cursor enter center target and then later figure out when the trajectory leaves that radius.

leoscholl:
> Correct me if I'm wrong, but these functions don't actually concatenate the trial segments. It looks to me like they extract the trial segments and place them all in a dataframe with their respective information. I tend to think of concatenation as when many trials become one continuous array.

That’s correct. We were thinking of the concatenation happening across recordings. Open to suggestions of a better name.

aopy/data/bmi3d.py
pavi-rajes: I agree with Leo, target_locations are useful to calculate movement error metrics. Target locations changes with task setup parameters like target size, distance, etc. It is useful to keep this in the data frame and it is a pain to add later for each row. 
 
Also, does it make sense to have task specific data processing functions? example: for 2D center out vs tracking task could have different functions, but this function should work for center out task irrespective of the number of targets. 
would be useful to get Katherine's input if you are thinking about generalizing for tracking task as well. But make this a future update. Let's get this function in copy now.

aopy/data/bmi3d.py
pavi-rajes: After trying out different iterations, I figured keeping track of event codes and event timestamps are an easy way to avoid mean of mean problems. This is a good intermediate step and we can use our existing behavior metrics functions on top of the event code arrays. It is also easier and flexible for data filtering through logical indexing.

pavi-rajes:
> Correct me if I'm wrong, but these functions don't actually concatenate the trial segments. It looks to me like they extract the trial segments and place them all in a dataframe with their respective information. I tend to think of concatenation as when many trials become one continuous array.

I agree too. concatenating trial segments is the first thing that comes to my mind too. how about concat_behavior_data() or something like that?

leoscholl:
Maybe “tabulate” instead of “concat”? tabulate_behavior_data() or tabulate_kinematic_segments()

aopy/data/bmi3d.py
rcanfield2: I think it is fine to keep the pull request as is, but I think the function name should be something like "concat_kin_trials" or "concat_trials_kindata" so the major restriction is obvious.

rcanfield2: I see now that this is addressed at in the comments below.

rcanfield2:
I like tabulate_behavior_data() because it can include all behavior data (rewards, eye tracking, optitrack, etc.) and not just the kinematic data.

rcanfield2 (approved):

PR #368, opened 2023-03-13
add test image

leoscholl:
generating some figures for the downsample function to see how it affects signals


pavi-rajes (approved):
changes look good.

PR #374, opened 2023-04-03
Tomo neuropixel functions

resatomo:
Load and sync functions for neuropixels.


aopy/precondition/base.py
```diff
@@ -393,6 +395,95 @@ def detect_spikes(spike_filt_data, samplerate, threshold, above_thresh=True,  wf
...
+        if ichunk == nchunk:
+            data = spike_filt_data[n_samples:,:]
+        else:
+            data = spike_filt_data[n_samples:n_samples+chunksize,:]
+
+        if above_thresh:
+            data_above_thresh_mask = data > threshold
+        else:        
+            data_above_thresh_mask = data < threshold
+
```
rcanfield2, line 440: We probably want this to output data either above or below the threshold. It probably isn't necessary to make this change in this pull request but we should open an issue detailing it.

rcanfield2: Actually, can we just use the absolute value of the signal and use a positive threshold?

rcanfield2:
I'm still working through this but we need to add the new packages (i.e. open ephys tools) to the docs. In the file "analyze/docs/source/conf.py" the new packages need to be added to the "autodoc_mock_imports" variable (line 17).

aopy/data/neuropixel.py
```diff
@@ -0,0 +1,195 @@
...
+def load_neuropixel_configuration(data_dir, data_folder, ex_idx = 0, port_number = 1):
+    '''
+    Get neuropixel probe information from xml condiguration files made by OpenEphys
+    This function also sorts x pos and y pos in the order of channel number
+    
+    Args:
+    data_dir (str): where to find the file
+    data_folder (str): the xml file that describes neuropixel probe configuration
+    ex_idx (int): experiment idx. This is usually 0.
+    port_number ()
```
rcanfield2, line 16: Missing description

leoscholl:
@resatomo @rcanfield2 docs isn't working, maybe need to add open_ephys to mock imports?

Issue #379, opened 2023-04-08
trial_align_times returns different number of data points at each trial


resatomo:
Returned indices have different number of data points at each trial.

leoscholl:
Can you explain why this is an issue? What timestamps are you trying to align that have a constant number of values for each trial?

resatomo:
I want to align synchronized timestamps and get indices to align lfp data of neuropixels. The following code is for aligning lfp data on laser-on times by using `trial_align_index` obtained by `trial_align_times` in aopy.

```
ntr = stim_on_times.shape[0]
nt = int((time_before+time_after)*FS)
nch = lfp.shape[1]

timestamp = lfp_data['sync_timestamp']
_, trial_align_index = aopy.preproc.trial_align_times(timestamp, stim_on_times, time_before, time_after, subtract=True)
trial_align_lfp = np.zeros((ntr,nt,nch))
for itr in range(ntr):
    trial_align_lfp[itr,:,:] = lfp[trial_align_index[itr],:]
```

This code looks okay, but actually doesn't work because `trial_align_index` has different number of samples at each trial. In this case, `trial_align_index` has (nt) samples in some trials, but it also has (nt+1) samples in other trials. So, in the last code `trial_align_lfp[itr,:,:] = lfp[trial_align_index[itr],:]`, we cannot insert the right to the left.

leoscholl:
why can't you use existing functions: `aopy.analysis.calc_erp` or `aopy.preproc.trial_align_data`?

resatomo:
This is related to what we talked about in the monkey meeting. Lfp data recorded by neuropixels is based on openephys timestamps and isn't synchronized to ecube. Also event times such as gocue and target onset are based on ecube timestamps. So, we cannot use `aopy.preproc.trial_align_data` and `aopy.analysis.calc_erp` which assumes data and event times are synchronized.

We might be able to synchronize lfp data itself by upsampling or downsampling, but it might be difficult because of memory issues or not integer sample rate. Instead of doing this, we already have synchronized timestamps. I want to align lfp data by aligning the synchronized timestamps and getting alignment indices.

leoscholl:
any update on getting `aopy.analysis.calc_erp` to work with non-integer sampling rate?

PR #381, opened 2023-04-08
Trial_align_times

resatomo:
#379 Added trial_aling_timestamps function. which aligns timestamps on triggered times and also returns indices to align other data. This function assumes timestamps whose intervals are equal (ex. [1,2,3,4,5,6,7,8]). 

There is a similar function (trial_align_times) in aopy. However, each trial split by this function often has different number of data points. But this function can deal with timestamps which is not equally spaced (ex. [1,5,8,9,15]). So I thought I should make trial_align_timestamps without deleting trial_align_times.

### DIFF
--- a/aopy/preproc/base.py
+++ b/aopy/preproc/base.py
@@ -382,8 +382,8 @@ def trial_align_times(timestamps, trigger_times, time_before, time_after, subtra
     
     Returns:
         tuple: tuple containing:
-            | **trial_aligned (ntrial, nt):** trial aligned timestamps
-            | **trial_indices (ntrial, nt):** indices into timestamps in the same shape as trial_aligned
+            | **trial_aligned (list of ntrial):** trial aligned timestamps
+            | **trial_indices (list of ntrial):** indices into timestamps in the same shape as trial_aligned
     '''
     trial_aligned = []
     trial_indices = []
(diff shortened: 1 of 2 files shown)


leoscholl:
can't you just make this matrix without doing any "alignment"? e.g.
```
samplerate = 10
time_before = 0.2
time_after = 0.2
trial_times = np.array([0.5, 1.7, 3.0]) # some made up trial times

def get_trial_aligned_timestamps(trial_times, samplerate, time_before, time_after):
    trial_timestamps = np.linspace(-time_before, time_after, int((time_after+time_before)*samplerate))
    timestamps = trial_times[np.newaxis,:] + trial_timestamps[:, np.newaxis]
    return timestamps

>>> get_trial_aligned_timestamps(trial_times, samplerate, time_before, time_after)
array([[0.3       , 1.5       , 2.8       ],
       [0.43333333, 1.63333333, 2.93333333],
       [0.56666667, 1.76666667, 3.06666667],
       [0.7       , 1.9       , 3.2       ]])
```

and to get indices
```
>>>(get_trial_aligned_timestamps(trial_times, samplerate, time_before, time_after)*samplerate).astype(int)
array([[ 3, 15, 28],
       [ 4, 16, 29],
       [ 5, 17, 30],
       [ 7, 19, 32]])
```
your version seems very complicated.

PR #385, opened 2023-04-27
Leo add rolling average

leoscholl:
Adding a function to do rolling averages where the output is the full size of the input but the invalid indices are replaced with the first and last valid values.


rcanfield2 (approved):
Looks good

PR #390, opened 2023-05-04
add memory default

leoscholl:


rcanfield2 (approved):
Looks good. I wonder if we should have a section of readthedocs about memory limits. It could include things about the hardware memory limits, inidividual limits, how to increase the individual memory limit if needed, etc.

PR #393, opened 2023-05-16
Update conf.py

rcanfield2:


leoscholl (commented):
can you also delete the memory functions from utils/base.py? they got moved to utils/memory.py

PR #401, opened 2023-06-08
Code for align dynamics using CCA 

pavi-rajes:
I implemented align dyanmics using CCA from scratch based on Galeego 2020 Long term stability of cortical dynamics paper. Please review and let me know if any other comments or tests would be helpful.

### DIFF
--- a/aopy/analysis/base.py
+++ b/aopy/analysis/base.py
@@ -268,7 +268,10 @@ def align_latent_dynamics(La, Lb, return_aligned_dynamics=False):
     CCs_unaligned = np.abs(np.diag(np.corrcoef(La, Lb)[:La.shape[0], La.shape[0]:])) # Pairwise correlations between rows of La and Lb
     CCs_aligned = S
 
-    return CCs_unaligned, CCs_aligned, La_tilde, Lb_tilde if return_aligned_dynamics else CCs_unaligned, CCs_aligned
+    if return_aligned_dynamics:
+        return CCs_unaligned, CCs_aligned, La_tilde.T, Lb_tilde.T
+    else:
+        return CCs_unaligned, CCs_aligned
 
 
 '''
--- a/aopy/analysis/base.py
+++ b/aopy/analysis/base.py
@@ -235,6 +235,42 @@ def calc_task_rel_dims(neural_data, kin_data, conc_proj_data=False):
     else:    
         return task_subspace.T, projected_data
 
+def align_latent_dynamics(La, Lb, return_aligned_dynamics=False):
+    """
+    Aligns latent dynamics using Canonical Correlation Analysis (CCA) and computes pairwise Pearson correlation for both aligned and unaligned dynamics.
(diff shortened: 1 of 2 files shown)


aopy/analysis/base.py
```diff
@@ -235,6 +235,45 @@ def calc_task_rel_dims(neural_data, kin_data, conc_proj_data=False):
...
         return task_subspace.T, projected_data
 
+def align_latent_dynamics(La, Lb, return_aligned_dynamics=False):
+    """
+    Aligns latent dynamics using Canonical Correlation Analysis (CCA) and computes pairwise Pearson correlation for both aligned and unaligned dynamics.
+    References: Gallego, J. A., Perich, M. G., Chowdhury, R. H., Solla, S. A. & Miller, L. E. Long-term stability of cortical population dynamics underlying consistent behavior. Nat Neurosci 23, 260–270 (2020).
+
+    Args:
+        La (ndarray): Latent dynamics of Dataset A with shape (m, n_timepoints). Usually first dimension is time, however Juancho's code has it as (m, n_t). Keeping it similar to his paper is easier for the computations below.
+        Lb (ndarray): Latent dynamics of Dataset B with shape (m, n_timepoints).
```
rcanfield2, line 245: I think to make things consistent for aopy we should keep the input data shapes the same other functions with time as the first dimension. Then the first couple lines in the function can transpose them and the transposed variables can be used throughout the rest of the function.

rcanfield2: Also, should the input data be raw neural data or data projected into a lower dimensional latent space?

aopy/analysis/base.py
```diff
@@ -235,6 +235,45 @@ def calc_task_rel_dims(neural_data, kin_data, conc_proj_data=False):
...
+
+    Returns:
+        CCs_unaligned (ndarray): Pairwise Pearson correlation between unaligned latent dynamics (La and Lb) with shape (m).
+        CCs_aligned (ndarray): Pairwise Pearson correlation between aligned latent dynamics (La_tilde and Lb_tilde) with shape (m).
```
rcanfield2, line 249: Also include the optional return args

tests/test_analysis.py
```diff
@@ -136,6 +137,49 @@ def test_get_unit_spiking_mean_variance(self):
...
+
+        # Call the align_latent_dynamics function
+        CCs_unaligned, CCs_aligned = aopy.analysis.align_latent_dynamics(La, Lb, False)
+
+        # Assert the shapes of the computed correlations
+        assert CCs_unaligned.shape == (10,)
+        assert CCs_aligned.shape == (10,)
+
+        # Assert that the pairwise correlation for aligned dynamics is approximately 0.99
+        assert np.allclose(CCs_aligned, 0.99, atol=0.01)
```
rcanfield2, line 179: It would be good to have a similar test with orthogonal data. I'm not exactly sure how to do this but maybe you run PCA on a dataset. Then, for La project data onto a few of the PCs, and for Lb project data onto the remaining PCs.

PR #406, opened 2023-06-21
Leo interp hotfix

leoscholl:
changed the way the `interp_timestamps2timeseries` function works because it was leading to a lot of nan data, which before was ignored but now is being fed into `filter_kinematics` among other things, which doesn't play well with nans. 

- changed default sampling rate everywhere to 1000
- changed extrapolation method to copy first value backwards and last value forwards by default
- added enforcement of monotonicity to the input timestamps
- updated figures that were affected by interpolation or sampling rate


aopy/preproc/base.py
```diff
@@ -179,15 +179,17 @@ def validate_measurements(expected_values, measured_values, diff_thr):
     corrected_values[diff_above_thr] = expected_values[diff_above_thr]
     return corrected_values, diff_above_thr
 
-def interp_timestamps2timeseries(timestamps, timestamp_values, samplerate=None, sampling_points=None, interp_kind='linear', extrap_values='extrapolate'):
+def interp_timestamps2timeseries(timestamps, timestamp_values, samplerate=None, sampling_points=None, interp_kind='linear', extrapolate=True):
```
rcanfield2, line 182: I think the default value for extrapolate should be False because it is an extension of what the function is designed to do. Extrapolation may happen that users are unaware of and could impact analysis/interpretations.

rcanfield2 (approved):
Look good.

aopy/preproc/base.py
leoscholl: Yeah I agree, was more trying to keep it consistent with the previous implementation

PR #410, opened 2023-06-26
Leo add db

leoscholl:
Adding db functions to aopy, would appreciate some testing & feedback on real data
- add bmi3d as a requirement for aopy
- lookup_sessions() function replaces db.get_task_entries(), has the same format but adds some functionality
- add lookup_flash_sessions(), lookup_mc_sessions(), lookup_tracking_sessions(), lookup_bmi_sessions() shortcuts
- lookup functions return a new wrapper BMI3DTaskEntry instead of working directly with the bmi3d TaskEntry model. can add future enhancements like being able to save links to preprocessed data, etc.
- add a couple wrapper utility functions


leoscholl:
hi pavi, have you had a chance to test these new functions out? one thing that I think might be missing is a way to lookup entries by their id number directly.

pavi-rajes:
> hi pavi, have you had a chance to test these new functions out? one thing that I think might be missing is a way to lookup entries by their id number directly.

Hi Leo, I have some issues installing this branch locally on crab-eating to test on real data. The problematic package is open-ephys-python-tools. here is the error I get: 
ERROR: Could not find a version that satisfies the requirement open-ephys-python-tools (from versions: 0.1.0, 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5)
ERROR: No matching distribution found for open-ephys-python-tools 

I liked the lookup functions in general. Some comments are: using some examples to filter sessions with ecube vs neuropixels recordings would be helpful. Also wondering is record_headstage is a feature?

leoscholl:
Something weird was going on with your pip. i ran
```
conda activate pavi_dev
pip cache purge
```
and then did
```
pip install -U pip
pip install open-ephys-python-tools
```
and it worked.

pavi-rajes:
I tested the dbfunctions on real data in the server in pavi_dev. The changes look great. database_example notebook helps understand what those functions do. Approving the pull request. 

**Some notes from my observations:** it looks like Django version should be <4.2 for the dbfunctions to work. For example, Django versions in the 'default' condo environment is 4.2.1 and this notebook doesn't run in that. Just something to note down. While I see that bmi3d requirements.txt specifies the version of Django to be installed, setup.py doesn't have any and I am not sure if setup.py uses requirements.txt to install packages. 
Here is the error message one would see: `NotSupportedError: MySQL 8 or later is required (found 5.7.40). error when Django version => 4.2 `

pavi-rajes (approved):

pavi-rajes:
Please check is tests/test_data.py is passing all tests on your end. 5 tests related to db failed in the server for me.

PR #411, opened 2023-06-27
kilosort functions

resatomo:
functions related to kilosort


resatomo:
I need to still change all functions so that we can concatenate trials within the same probe configuration, subject, and port.

PR #416, opened 2023-07-07
Calc waveforms from kilosort output data

resatomo:
Get waveforms and unit positions from raw data in neuropixels, using spike indices of each unit obtained by kilosort

### DIFF
--- a/aopy/precondition/base.py
+++ b/aopy/precondition/base.py
@@ -601,6 +601,7 @@ def bin_spikes(data, fs, bin_width):
 def calc_ks_waveforms(raw_data, sample_rate, spike_times_unit, templates, channel_pos, waveforms_nch=10, time_before=1000., time_after=1000.):
     '''
     Calculate waveforms, waveform channels, and positions of units, using templates from kilosort
+    This function does not account for drift correction
     
     args:
         raw_data (nt,nch): time series neural data to detect spikes and extract waveforms from.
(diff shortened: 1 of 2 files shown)


rcanfield2:
Does this account for drift correction? It doesn't look like it does

resatomo:
In my understanding, spike times and their clusters already result from drift correction. It computes drift map based on spike amplitudes, and correct drift at each batch. Then it computes spikes times again.

resatomo:
I got what you meant now. Spike times are affected by drift correction. But channel mapping files are static over time. Do you think we need waveforms even after drift?

aopy/precondition/base.py
```diff
@@ -598,6 +598,62 @@ def bin_spikes(data, fs, bin_width):
     binned_spikes = binned_spikes/bin_width # convert from [spikes/bin] to [spikes/s]    
     return binned_spikes
 
+def calc_ks_waveforms(raw_data, sample_rate, spike_times_unit, templates, channel_pos, waveforms_nch=10, time_before=1000., time_after=1000.):
+    '''
+    Calculate waveforms, waveform channels, and positions of units, using templates from kilosort
+
```
rcanfield2, line 604: Can you add a comment here that this function does not account for drift correction?

rcanfield2 (approved):

PR #418, opened 2023-07-18
Update base.py

pavi-rajes:
updated assertion to check normalized new_axis3D vector against output_axis.Otherwise function doesn't work when new_axis is not a unit vector.


leoscholl (approved):
Thanks!

PR #428, opened 2023-08-07
Get saccade labels

resatomo:
Added a function to get which targets a saccade is for and event code passed just before the saccade occur. Let me know if you have a good function name.


leoscholl:
Cool looks like it works. Could you maybe separate some of the inner code into its own helper functions, e.g.  ‘get_saccade_vectors’ that inputs eye pos and saccade times and outputs the start and end positions of each saccade, and maybe ‘get_saccade_events’ can input saccade times and event times and output the event for each saccade? It would clean up your function a lot but also probably those will be useful for Katherine or others who aren’t using center out task. 

I wonder if these functions might be better suited to somewhere like aopy.postproc.eye? Or aopy.postproc.target_acq? You should probably append ‘_center_out’ or ‘_target_acq’ to your function also.

leoscholl (approved):
I asked chatgpt to vectorize your code. i think it is a lot faster now.

PR #431, opened 2023-08-07
Leo gradient plots

leoscholl:
adding functions to plot timeseries and 2d/3d trajectories with a gradient


docs/source/_images/gradient_trajectories_simple.png
katherineperks, line 1: The way I understand the code, would the trajectory with 5 points (4 segments) be made up of 4 colors?

docs/source/_images/gradient_trajectories_3d.png
katherineperks, line 1: Just a note that if this plot is included as an example in the docs, hand movements front-back are plotted on the x-axis, up-down on y-axis, and left-right on z.

katherineperks (commented):
Besides the comment I left about # segments matching # colors, everything else makes sense

docs/source/_images/gradient_trajectories_3d.png
katherineperks: Nothing wrong with the code, but I don't understand how these bounds work because the trajectory values are centered on [0, 70, 36] ?

docs/source/_images/gradient_trajectories_simple.png
leoscholl: you always choose the number of colors. but yes, 4 is the maximum number of colors you could choose for such a trajectory. you could choose less, like 3, in which case the last segment would get the same color as the 3rd segment. i added some comments and made a check that the n_colors is smaller than the number of timepoints.  

it might be nicer to have a way to automatically choose the right number of colors so that you get a smooth trajectory without it taking a really long time to plot. but in my tests 100 colors seemed to work all the time for cursor trajectories, so to keep it simple i just went with that. let me know if it makes sense.

docs/source/_images/gradient_trajectories_3d.png
leoscholl: hmm yeah that's a good point, we'll have to update this once we have some code to map the optitrack coordinates into something more sensible. 

i played around with the 3d plot, i think they were out in front of the screen. better bounds seem to be [-10,0,60,70,20,40] which makes more sense given the optitrack coordinates!

katherineperks (approved):
Looks good! Thanks for adding clarifying comments

PR #432, opened 2023-08-08
add tabulate_kinematic_data

leoscholl:
some ideas on how to better deal with tabulating kinematics and other large data across many recordings

- the get_kinematics() function still needs work to include things katherine is working on, and to apply transformations so we can get hand position in screen coordinates, e.g. 

- the tabulate_kinematic_data() function is my idea on how we might load kinematics, lfp, spike times, etc. from a dataframe. let me know what you think. it uses caching in a few places to avoid loading the same file over and over again.


leoscholl:
just a couple more notes to flesh out what i'm thinking.. to-do:
- add some functions to define a "trial" for various tasks that input event_codes and event_times and output a start and end time for that trial, e.g. center_out_reach trial might start with code 32 and end with 80-89. 
- modify tabulate_behavior_data to not include any kinematics by default

The workflow would then look like:
- make a big df of all the trials you want to analyze using tabulate_behavior_data
- use the trial defn functions to get start and end times of each trial, specific to the analysis you're doing (might be shorter than the entire length of the trial, or maybe you want just the beginning of the trial plus/minus some fixed length)
- tabulate_kinematic_data (and we can write tabulate_lfp_data) to specify the data you want to analyze, potentially also saved into some sort of memory-efficient file

leoscholl:
i added tabulate_ts_data which should work with ecog neural data, along with helper functions to get timeseries segments from hdf files. could you both please take a look and leave comments? thanks!

aopy/data/bmi3d.py
katherineperks, line 1: Could including target_idx & target_location be part of tabulate_behavior_data_center_out? I don't think these fields make sense for any non-target-capture task.

katherineperks: I would like get_kinematics & get_interp_kinematics to be able to return 3 other datatypes which are specific to tracking task:

katherineperks:
I like the workflow! The functions make a lot of sense to me. For this workflow to be implemented in the tracking task analysis pipeline, I will need to add a few things but it seems ready to go for center out.

tests/test_data.py
```diff
@@ -582,6 +572,48 @@ def test_tabulate_behavior_data_center_out(self):
...
+    def test_tabulate_kinematic_data(self):
+        subjects = [self.subject, self.subject]
+        ids = [self.te_id, self.te_id]
+        dates = [self.date, self.date]
+
+        df = tabulate_behavior_data_center_out(write_dir, subjects, ids, dates, df=None, 
+                                               include_center_target=True,
+                                               include_handdata=False, include_eyedata=False)
+
+        start_times = [t[0] for t in df['event_times']] # maybe we could update the df to have start and end times directly
```
katherineperks, line 584: I think it's more flexible for the df to have all events & event_times and you specify the start and end times before calling tabulate_kinematic_data. For example in tracking task, the first and last event of each segment are not really the start and end times because I have to exclude the ramp period. If I wanted to include the ramp period in future analyses, it'd be faster to skip this step than re-generate the df with correct start & end times.

aopy/data/bmi3d.py
katherineperks: If I want to use the result from tabulate_behavior_data to perform spectral analysis, df will need the following fields:
- ref_traj
- dis_traj
- input_traj
- ref_freq_inds
- dis_freq_inds
- perturbation_rotation (using the incremental_rotation feature means this changes over trials within a session)

My current workflow is sort of like tabulate_behavior_data where within the for loop over sessions, I perform spectral analysis and include freq domain signals in the tabulation - rather than tabulating all time domain signals and then doing spectral analysis after.

leoscholl:
Thanks for the comments @katherineperks! 

- Agree about target_idx and target_location. It can go in the center out wrapper. 
- For things like reference and disturbance trajectories, it would be better if the tabulate_kinematics() function handled that.
- For metadata like perturbation rotation that changes across trials, similar to target_idx and target_location, you'll want to put those in a wrapper function around tabulate_behavior_data(). 
- For simple metadata like target_radius that's consistent for a given recording, it would be cool if there was an argument to tabulate_behavior_data() that let you add whatever metadata you want.

leoscholl:
made some minor improvements and reorganized a bit:
- moved `target_idx` and `target_location` from tabulate_behavior_data() to tabulate_center_out
- added a `metadata` option to tabulate_behavior_data() to include various metadata from the experiment
- modified get_interp_kinematics() and get_kinematics() to grab any task data, so they will work with reference and disturbance trajectories, perhaps.
- removed the old get_kinematic_segments() and other duplicate functions from aopy.postproc

leoscholl:
remaining to-do (can be separate PRs from this one):
- add some functions to define a "trial" for various tasks that input event_codes and event_times and output a start and end time for that trial, e.g. center_out_reach trial might start with code 32 and end with 80-89.
- add a tabulate_tracking wrapper for tracking task-specific info like reference indices, perturbation rotations, etc.
- add some sort of trial identity to tabulated data, consistent across dataframes
- write some code to save big tabulated data (lfp trials, e.g.) into files that can be accessed in memory-efficient ways
- write functions to tabulate experiments across learning series (ABA manual control or tracking rotation experiments, e.g.)

pavi-rajes (approved):
changes look good. test_data failed on those functions that require bmi3d on my end. Please check once if all tests pass on your end before merging.

pavi-rajes:
Just some notes: For center out, I am adding target_radius as a column in the df. Current code allows it through providing a metadata key as an input to the tabulate_behavior_data_center_out function, which is cool.

pavi-rajes:
would it make sense to have a separate center out.py and tracking task.py file in postproc or somewhere to move task specific wrappers functions to ?

Issue #412, opened 2023-06-29
load bmi3d center_out task codes function


leoscholl:
something like:
def load_center_out_task_codes():
    config_dir = files('aopy').joinpath('config')
    params_file = as_file(config_dir.joinpath('task_codes.yaml'))
    with params_file as f:
        task_codes = yaml_read(f)[0]
    return task_codes

leoscholl:
also a similar function for channel mapping files would be useful, to load the file so that it can be used in map_ch2acq, etc.

PR #433, opened 2023-08-08
add helper fn for task codes

leoscholl:
fixes issue #412


leoscholl:
work in progress. i will add another helper for channel mappings, too

PR #435, opened 2023-08-10
change criteria

leoscholl:
minor fix to proc_eye_day. for some reason the flash experiments sometimes yield a 1.0 correlation, even though there's no cursor data? not sure.

i still want to make a version of this function that doesn't modify the preprocessed files, but this works for now.


katherineperks (approved):
Looks good. Will follow up on improving the eye preproc function and the function that applies calibration on days

PR #436, opened 2023-08-10
update rolling average

leoscholl:
made some changes including:
- allow for 2d arrays, e.g. (nt, 2) cursor kinematics
- fix a problem where if the kernel size was bigger than the input, the output would be the wrong shape
- add a mode where the invalid output is 'nan' instead of copying the first and last valid output
- updated the test functions


aopy/analysis/base.py
```diff
@@ -441,27 +441,44 @@ def calc_sem(data, axis=None):
...
 
-def calc_rolling_average(data, window_size=11):
+def calc_rolling_average(data, window_size=11, mode='copy'):
     """
-    Computes the rolling average of a 1D array using a convolutional kernel. The
-    first and last valid datapoint (fully overlapping with the kernel) are copied
-    backwards and forwards, respectively, such that the size of the output is the
-    same as the size of the input.
+    Computes the rolling average of a 1- or 2-D array using a convolutional kernel. 
+    The rolling average is always applied along the first axis of the array.
```
rcanfield2, line 447: This should work for an n-D array right?

leoscholl: Yeah probably, but untested

PR #437, opened 2023-08-10
add options to sessions_by_trial

leoscholl:
work in progress


pavi-rajes (approved):
changes look good. In the future revisions of this code, we can add additional functionalities like, plotting by session if required. 
Also, there is a way to use inbuilt rolling.mean() function in data frame to get the rolling success rates. for example: df.groupby(['mapping', 'session']).rolling(15).mean().reset_index() does generate a similar plot without the edge effects. More on this could be a follow up pull request. Approving this version since it generates the required plot. :)

PR #443, opened 2023-08-21
compute tuning significance

rcanfield2:


aopy/analysis/tuning.py
```diff
@@ -63,26 +64,37 @@ def get_preferred_direction(b1, b2):
     return pd
 
 
-def get_mean_fr_per_direction(data, target_dir):
+def get_mean_fr_per_condition(data, condition_labels, return_significance=False):
     '''
+    This function computes the average activity within a time window for each feature and trial.
```
resatomo, line 69: It's better to remove 'within a time window' because it doesn't use the time window in computing average activity?

PR #445, opened 2023-08-24
add function to compute an array of RGB values that represent a gradient.

rcanfield2:


leoscholl (approved):
This looks good. I'll just point out that you could also use seaborn's built-in color_palette function to accomplish this: 

```
colors = sns.color_palette("flare", n_colors) 
```

will yield a list of rgb colors belonging to an existing colormap, or
```
colors = sns.color_palette("blend:red,green", n_colors) 
```
for a gradient between any two valid color strings. could we propose encouraging folks to use the more powerful seaborn package instead?

PR #446, opened 2023-08-24
Update spectrograms

leoscholl:
- add calc_stft_tfr function using short-time fourier transform
- change mike's sgram function into calc_mt_tsa_tfr compatible with calc_mt_tfr
- update calc_mt_tfr to vectorize computing ffts over each taper
- add unified tests for above


resatomo (approved):
Looks good

PR #450, opened 2023-09-01
add prepare erp

leoscholl:
to make sure erp is mean zero in null condition!


resatomo (approved):
Looks good. But I think it's better to change code in the future so that this process can be done for different shape of null condition?

PR #452, opened 2023-09-11
spectrogram alignment

leoscholl:
- corrected the number of windows
- removed the (time - n/2) alignment, now data is aligned on center by default
- added alignment images to the docstring


leoscholl:
i'm still confused about alignment to left/right/center

PR #455, opened 2023-09-20
always include eye closed mask

leoscholl:
minor fix to always save the eye closed mask. it was getting thrown away for data without calibration


katherineperks (approved):
Looks good!

PR #458, opened 2023-09-29
tabulate laser data

leoscholl:
work in progress


leoscholl:
waiting for #454 to be approved before the tests will make sense

katherineperks (approved):
Looks good. I can't check if the image generated by your test is the same as `laser_aligned_sensor_debug.png` because I'm not running the tests on the lab server!

leoscholl:
> I can't check if the image generated by your test is the same as `laser_aligned_sensor_debug.png` because I'm not running the tests on the lab server!

It should be in analyze/tests/tmp/ wherever you run the tests!

PR #461, opened 2023-10-17
add  importlib_resources backport for python 3.8 and below

sijia66:


leoscholl (approved):

sijia66:
if everyone else is fine, we don't have to make it a general rule. I will patch these as needed.

PR #462, opened 2023-10-17
Katherine tabulate tracking task

katherineperks:
A tracking task-specific wrapper function for tabulate_behavior_data(). I may add more columns of info to the df, like the frequency content of reference & disturbance for each trial


aopy/data/bmi3d.py
```diff
@@ -889,6 +890,84 @@ def tabulate_behavior_data_center_out(preproc_dir, subjects, ids, dates, metadat
...
+        task_codes = yaml_read(f)[0]
+    trial_end_codes = [task_codes['REWARD'], task_codes['TIMEOUT_PENALTY'], task_codes['HOLD_PENALTY'], task_codes['OTHER_PENALTY']]
+    reward_codes = [task_codes['REWARD']]
+    
+    if include_before_initiation:
+        trial_start_codes = [task_codes['CENTER_TARGET_ON']]
+        penalty_codes = [task_codes['TIMEOUT_PENALTY'], task_codes['HOLD_PENALTY'], task_codes['OTHER_PENALTY']]
+    else:
+        trial_start_codes = [task_codes['TRIAL_START']]
+        penalty_codes = [task_codes['HOLD_PENALTY'], task_codes['OTHER_PENALTY']]
```
leoscholl, line 930: This is fine and similar to the way i set up the center_out function. But i've been meaning to simplify it so that it always includes the entire trial (from when the center target turns on until the trial end code) so that the dataframes are more consistent and you never have to remake them. I suggest you do the same at some point. It just requires that you be more thoughtful about selecting trial start and end times later, and we should write some functions that take in one trial worth of event codes and spit out the indices that define various forms of sub-trial, like a center-out reach or a tracking-in period, etc.

aopy/data/bmi3d.py
```diff
@@ -889,6 +890,84 @@ def tabulate_behavior_data_center_out(preproc_dir, subjects, ids, dates, metadat
...
+    
+    # Concatenate base trial data
+    metadata.append('sequence_params')
+    new_df = tabulate_behavior_data(
+        preproc_dir, subjects, ids, dates, trial_start_codes, trial_end_codes, 
+        reward_codes, penalty_codes, metadata, df=None)
+    
+    # Add trajectory timing info
+    trajectory_times = [
+        [new_df.iloc[i]['event_times'][np.where(code==task_codes['CURSOR_ENTER_TARGET'])[0][0]], new_df.iloc[i]['event_times'][-1]] # HOLD complete, REWARD/OTHER_PENALTY
```
leoscholl, line 940: I think this should be its own function

leoscholl:
Hey i recently merged some changes to get_interp_kinematics, so i updated them here as well

aopy/data/bmi3d.py
```diff
@@ -13,6 +13,8 @@
 import h5py
 import tables
 import pandas as pd
+import json
+import sympy
```
leoscholl, line 17: you will need to add sympy to /docs/source/conf.py under 'mock_imports' as well so that the docs don't get messed up. see https://analyze.readthedocs.io/en/latest/code-guidelines.html?highlight=mock#adding-new-packages

aopy/data/bmi3d.py
```diff
@@ -832,6 +835,46 @@ def get_target_locations(preproc_dir, subject, te_id, date, target_indices):
...
+        trial_d_idx = np.array([odd_idx, even_idx]*params['ntrials'], dtype='object')
+    elif order == 1:
+        trial_r_idx = np.array([odd_idx, even_idx]*params['ntrials'], dtype='object')
+        trial_d_idx = np.array([even_idx, odd_idx]*params['ntrials'], dtype='object')
+        
+    # get trajectory generator index used for each trial
+    trial_start_codes = [task_codes['CENTER_TARGET_ON']]
+    trial_end_codes = [task_codes['REWARD'], task_codes['OTHER_PENALTY'], 
+                       task_codes['HOLD_PENALTY'], task_codes['TIMEOUT_PENALTY']] # using TRIAL_END may not pass assertion
+    gen_trajectories, _ = get_kinematic_segments(preproc_dir, subject, te_id, date,
```
leoscholl, line 867: i think it might make more sense to use the get_trial_segments() function to segment the `data['task']['gen_idx']` vector into trials using `data['bmi3d_events']['time']` and `data['bmi3d_events']['code'] rather than using real-time get_kinematic_segments() wrapper, which uses `data['events']['timestamp']` and `data['events']['code']` recorded by the ecube. That way the events are the same, and therefore the segments will be the same, but there's no interpolation.

aopy/data/bmi3d.py
```diff
@@ -832,6 +835,46 @@ def get_target_locations(preproc_dir, subject, te_id, date, target_indices):
...
+    gen_trajectories, _ = get_kinematic_segments(preproc_dir, subject, te_id, date, 
+                                                 trial_start_codes, trial_end_codes, datatype='gen_idx', 
+                                                 samplerate=metadata['fps'], filter_kinematics=False)
+    assert np.all([int(gen[0])==int(gen[-1]) for gen in gen_trajectories])
+    gen_idx = [int(gen[0]) for gen in gen_trajectories]
+
+    # use generator index to get reference & disturbance frequencies for each trial
+    freq_r = [primes[np.array(idx, dtype=int)]/base_period for idx in trial_r_idx[gen_idx]]
+    freq_d = [primes[np.array(idx, dtype=int)]/base_period for idx in trial_d_idx[gen_idx]]
+    return freq_r, freq_d
```
leoscholl, line 876: could all of this be moved into a function inside `../preproc/bmi3d.py`? This is no longer a `data` function

aopy/data/bmi3d.py
```diff
@@ -979,6 +1022,106 @@ def tabulate_behavior_data_center_out(preproc_dir, subjects, ids, dates, metadat
...
+        df (DataFrame, optional): pandas DataFrame object to append. Defaults to None.
+
+    Returns:
+        pd.DataFrame: pandas DataFrame containing the concatenated trial data
+    '''
+    # Use default "trial" definition
+    config_dir = files('aopy').joinpath('config')
+    params_file = as_file(config_dir.joinpath('task_codes.yaml'))
+    with params_file as f:
+        task_codes = yaml_read(f)[0]
```
leoscholl, line 1047: you can use the new `load_bmi3d_task_codes()` function

aopy/data/bmi3d.py
katherineperks: Does it make sense to just move the entire `get_trajectory_frequencies()` function to preproc/bmi3d.py?

leoscholl:
coming along nicely. you should write some tests!

aopy/data/bmi3d.py
leoscholl: not the entire thing because it loads the file from preproc/subject/etc. the preproc functions generally load exp_data, exp_metadata, etc.

aopy/preproc/bmi3d.py
```diff
@@ -723,4 +725,55 @@ def get_target_events(exp_data, exp_metadata):
...
+
+    Args:
+        data (dict): A dictionary containing the experiment data.
+        metadata (dict): A dictionary containing the experiment metadata.
+
+    Returns:
+        tuple: Tuple containing:
+            | **freq_r (list of arrays):** (ntrial) list of (nfreq,) frequencies used to generate reference trajectory
+            | **freq_d (list of arrays):** (ntrial) list of (nfreq,) frequencies used to generate disturbance trajectory
+    '''
```
leoscholl, line 744: can you add the examples and images from your test function here? that way they will show up in readthedocs. see https://analyze.readthedocs.io/en/latest/_modules/aopy/analysis/base.html#calc_cwt_tfr for an example

leoscholl (approved):
looks good!

PR #470, opened 2023-10-23
Add trial definition function

leoscholl:
- moved get_target_events() to preproc
- had to change some imports to avoid circularity
- added a "load_bmi3d_task_codes" helper function
- added "locate_trial_segment" function to get the event times for a given trial segment, currently center out only


tests/test_data.py
katherineperks, line 1: Is it normal that all the tests under DatabaseTests fail?

katherineperks (approved):
Looks good!

aopy/data/bmi3d.py
```diff
@@ -940,11 +892,8 @@ def tabulate_behavior_data_center_out(preproc_dir, subjects, ids, dates, metadat
...
-    config_dir = files('aopy').joinpath('config')
-    params_file = as_file(config_dir.joinpath('task_codes.yaml'))
-    with params_file as f:
-        task_codes = yaml_read(f)[0]
-    trial_end_codes = [task_codes['REWARD'], task_codes['TRIAL_END']]
+    task_codes = load_bmi3d_task_codes()
+    trial_end_codes = [task_codes['TRIAL_END']]
     reward_codes = [task_codes['REWARD']]
     
     if include_center_target:
```
katherineperks, line 899: Wouldn't you want to always include the full trial, i.e. from when cursor enters center target to trial end? Otherwise you won't be able to fill in some of the trial segment timing columns

leoscholl: Yes i missed that, thanks

PR #471, opened 2023-10-26
add subplots with letter numbering

sijia66:
a function that augments plt.subplots by adding letter numbering to the subplots.


pavi-rajes (approved):
looking good. Cool feature! All tests passing on my end. 

One suggestion (not needed for merging PR) would be to show the generated test figure. You could save the figure in the tmp folder a.k.a write_dir using savefig(write_dir, <filename>)

leoscholl:
i added the example image to the docstring, it really helps to be able to see the example figure on readthedocs.

could you also update the docstring format to match the style guide? https://analyze.readthedocs.io/en/latest/documentation-guidelines.html

PR #481, opened 2023-11-13
add tabulate flash

leoscholl:
- add the option to use event names instead of event code for alignment in tabulate_behavior_data
- add tabulate_behavior_data_flash

haven't added tests yet


katherineperks (approved):
Looks good. Can you update the docstring for `tabulate_behavior_data_flash()` before merging?

PR #482, opened 2023-11-13
add reset_color

leoscholl:
quick utility so i don't have to remember how to reset the color cycler


sijia66 (approved):
Documentation is very helpful to understand the complicated one line function!

PR #491, opened 2023-12-28
maybe fix loading corrupted data

leoscholl:


leoscholl:
code to delete task data from an existing hdf file:

```
# Make a dummy HDF file by copying an existing HDF file from the same experiment
import shutil
files['hdf'] = 'dummy_hdf_rig1_center_out.hdf'
shutil.copy(os.path.join(data_dir, 'beig20230109_15_te7977.hdf'), os.path.join(data_dir, files['hdf']))
with tables.open_file(os.path.join(data_dir, files['hdf']), 'r+') as f:
    f.get_node('/task').truncate(0)
```

tests/test_preproc.py
```diff
@@ -903,6 +904,27 @@ def test_parse_bmi3d_v13(self):
...
+
+        # Test if a dummy HDF file is provided
+        files['hdf'] = 'dummy_hdf_rig1_v13.hdf'
+
+        # Make a dummy HDF file by copying an existing HDF file from the same experiment
+        import shutil
+        shutil.copy(os.path.join(data_dir, 'beig20230109_15_te7977.hdf'), os.path.join(data_dir, files['hdf']))
+        with tables.open_file(os.path.join(data_dir, files['hdf']), 'r+') as f:
+            f.get_node('/task').truncate(0)
+
```
rcanfield2, line 922: Do we need to make a wrapper to do this?

rcanfield2 (approved):

tests/test_preproc.py
leoscholl: Yeah something that wraps around proc_exp() maybe?

rcanfield2: yeah, something like proc_exp_new_hdf()

PR #494, opened 2024-01-10
add default false to caching functions

leoscholl:
fixing my assumption that hdf data is static


rcanfield2 (approved):
Looks good, I'm not sure how to test that removing the cache works until I try it in a notebook.

PR #497, opened 2024-01-19
Update to tabulate tracking task

katherineperks:
Some updates to the tracking task wrapper so that it can handle older data in which certain params were not saved. Also add optional argument to `tabulate_behavior_data()` to return the list of task entries that could not be loaded.


leoscholl (approved):
Looks good to me!

PR #498, opened 2024-01-19
Leo update eye calibration

leoscholl:
slight changes to the eye calibration function 
- ignore failed trials 
- consider any peripheral target instead of just the first peripheral target in a trial


katherineperks:
Looks good, can you add something in metadata to keep track of which version of eye calibration was used?

katherineperks (approved):
Looks good!

PR #499, opened 2024-01-19
Update laser find_stim_times

leoscholl:
- calculate laser power in addition to gain
- fix a bug where laser gain was zero
- separate the pipeline for old laser data from new laser data
- integrate laser events into bmi3d events (work in progress)


aopy/preproc/laser.py
```diff
@@ -0,0 +1,171 @@
...
+            plt.plot(gain, powers)
+            plt.xlabel('gain')
+            plt.ylabel('power (mW)')
+
+        .. image:: _images/calibrate_laser_gain.png
+    '''
+    config_dir = files('aopy').joinpath('config')
+    with as_file(config_dir.joinpath(calibration_file)) as f:
+        yaml_data = aodata.yaml_read(f)
+    voltages = np.interp(gain, yaml_data['gain'], yaml_data['power'])
```
rcanfield2, line 138: This is probably fine for now but since the relationship between power and gain is nonlinear we should probably fit an exponential to the measurements and then calculate the power from that.

aopy/preproc/laser.py
```diff
@@ -0,0 +1,171 @@
...
+
+    Returns:
+        (ntrial,) float array: laser power in mW
+    '''
+    config_dir = files('aopy').joinpath('config')
+    with as_file(config_dir.joinpath(calibration_file)) as f:
+        yaml_data = aodata.yaml_read(f)
+    
+    # Use interpolation in case calibration doesn't include 0. or 1. gain setting
+    voltages = np.interp([0.,1.], yaml_data['gain'], yaml_data['power'])
```
rcanfield2, line 165: same as comment before.

rcanfield2 (approved):

PR #506, opened 2024-01-26
db decoder functions

leoscholl:
add db functions for fetching and loading decoders


aopy/data/db.py
```diff
@@ -114,6 +115,34 @@ def lookup_bmi_sessions(bmi_task_name='bmi control', **kwargs):
...
+
+    Returns:
+        list: list of Decoder records matching the query
+    '''
+    if id is not None and isinstance(id, list):
+        kwargs['id__in'] = id
+    elif id is not None:
+        kwargs['id'] = id
+    if parent_id is not None:
+        kwargs['entry_id'] = parent_id
```
rcanfield2, line 141: Do you need to do the same condition if parent_id is or is not a list?

aopy/data/db.py
```diff
@@ -414,6 +443,96 @@ def get_db_object(self):
         '''
         return self.record
 
+class BMI3DDecoder():
+    '''
+    Wrapper for BMI3D Decoder objects
+    '''
+
```
rcanfield2, line 450: Can you add a little more information about what this class does with BMI3D decoder objects?

aopy/data/db.py
```diff
@@ -414,6 +443,96 @@ def get_db_object(self):
...
+        Constructor
+
+        Args:
+            task_entry (TaskEntry): a database task entry from BMI3D
+            dbname (str, optional): name of the database to connect to. Defaults to 'default'.
+        '''
+        self.dbname = dbname
+        self.record = decoder
+        self.id = self.record.id
+        self.parent = lookup_sessions(id=decoder.entry_id)[0]
```
rcanfield2, line 462: Can decoder.entry_id be empty or None?

aopy/data/db.py
```diff
@@ -414,6 +443,96 @@ def get_db_object(self):
...
+        self.name = self.record.name
+        
+    def __repr__(self):
+        '''
+        Call __repr__ on the database record
+
+        Returns:
+            str: repr
+        '''
+        return f"Decoder[{self.id}:{self.name}] trained from {repr(self.parent)}"
```
rcanfield2, line 474: Same as above comment if applicable

rcanfield2 (approved):

aopy/data/db.py
leoscholl: i don't think it makes as much sense to lookup decoders from multiple parents..?

aopy/data/db.py
leoscholl: it's a wrapper :)

aopy/data/db.py
leoscholl: good question, i don't think so but it can't hurt to check

PR #508, opened 2024-01-27
get mean neighboring channels

resatomo:
Average ECoG data across neighboring channels based on distance.

### DIFF
--- a/aopy/analysis/base.py
+++ b/aopy/analysis/base.py
@@ -1541,3 +1541,42 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
     NCC[nan_idx1] = np.nan
     
     return NCC, shifts
+
+def get_neighboring_channel_index(elec_pos, distance=0.75):
+    '''
+    Get neiboring channel index based on distance
+    
+    Args:
+        elec_pos (nch, 2): x, y position of each channel
+        distance (float): define neighboring channels based on this distance. 0.75 is for 4 neighboring channels and 1.3 is for 9 channels.
+        
+    Returns:
+        neighboring_idx (nch, nch): boolean array to show neighboring channels for each channel
+    '''
+    dist = utils.calc_euclid_dist_mat(elec_pos)
+    neighboring_idx = dist <= distance
+        
+    return neighboring_idx
+
+def get_mean_neighboring_channels(data, elec_pos, distance=0.75):
+    '''
+    Calculates mean of data across neighboring channels based on distance
+    
+    Args:
+        data (nch): neural data
+        elec_pos (nch, 2): x, y position of each channel
+        distance (float): define neighboring channels based on this distance. 0.75 is for 4 neighboring channels and 1.3 is for 9 channels.
+        
+    Returns:
+        m_data (nch): Averaged data across neighboring channels
+    '''
+
+    dist = utils.calc_euclid_dist_mat(elec_pos)
+    nch = dist.shape[0]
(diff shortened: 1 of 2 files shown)


aopy/analysis/base.py
```diff
@@ -1621,3 +1621,42 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
     NCC[nan_idx1] = np.nan
     
     return NCC, shifts
+
+def get_neighboring_channel_index(elec_pos, distance=0.75):
```
leoscholl, line 1625: what do you think about having this function take "ch_idx" as an argument, and only return neighboring_idx[ch_idx], instead of the whole matrix?
i think it might be more useful that way? 
probably it will be slower, so in get_mean_neighboring_channels() you might not use this function

aopy/analysis/base.py
```diff
@@ -1621,3 +1621,42 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+    return neighboring_idx
+
+def get_mean_neighboring_channels(data, elec_pos, distance=0.75):
+    '''
+    Calculates mean of data across neighboring channels based on distance
+    
+    Args:
+        data (nch): neural data
+        elec_pos (nch, 2): x, y position of each channel
+        distance (float): define neighboring channels based on this distance. 0.75 is for 4 neighboring channels and 1.3 is for 9 channels.
```
leoscholl, line 1648: 0.75 and 1.3 only applies to ECoG244 array

aopy/analysis/base.py
```diff
@@ -1621,3 +1621,42 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+        
+    Returns:
+        neighboring_idx (nch, nch): boolean array to show neighboring channels for each channel
+    '''
+    dist = utils.calc_euclid_dist_mat(elec_pos)
+    neighboring_idx = dist <= distance
+        
+    return neighboring_idx
+
+def get_mean_neighboring_channels(data, elec_pos, distance=0.75):
```
leoscholl, line 1641: i don't want to derail this function -- it's fine and you can push it -- but just a note for the future: we should really make a spatial smoothing function that takes an arbitrary 2d kernel and smooths the data that way, rather than only doing average and only limited to a radius

tests/test_analysis.py
```diff
@@ -209,6 +209,17 @@ def test_calc_task_rel_dims(self):
...
+    def test_mean_neighboring_channels(self):
+        elec_pos_test = np.array([[1,1],[2,1],[3,1],[1,2],[2,2],[3,2],[1,3],[2,3],[3,3]])
+        neighboring_idx = aopy.analysis.get_neighboring_channel_index(elec_pos_test, distance=1)
+        self.assertTrue(np.all(neighboring_idx.T == neighboring_idx))
+        self.assertTrue(neighboring_idx[:,4].sum()==5)
+        
+        data_test = np.arange(1,10)
+        m_data_test = aopy.analysis.get_mean_neighboring_channels(data_test, elec_pos_test, distance=1)
+        self.assertTrue(m_data_test[0] == np.sum(data_test[0]+data_test[1]+data_test[3])/3)
+        self.assertTrue(m_data_test[4] == np.sum(data_test[1]+data_test[3]+data_test[4]+data_test[5]+data_test[7])/5)
```
leoscholl, line 221: it would be great to add a couple more lines to get a spatial map of the test data before and after smoothing and make a plot of before/after. then you could include the plot and the example code in the function docstring as well.

PR #509, opened 2024-01-29
remove default filtering in get_interp_kinematics

leoscholl:
we can use preproc=aopy.precondition.filter_kinematics() instead if needed


aopy/precondition/base.py
```diff
@@ -758,11 +758,13 @@ def filter_lfp(broadband_data, broadband_samplerate, lfp_samplerate=1000., low_c
         buttord (int, optional): order for butterworth low-pass filter. Defaults to 4.
 
     Returns:
-        (nt', ...): lfp data
+        tuple: tuple containing:
+        | lfp_data **(nt', ...):** downsampled filtered lfp data
```
resatomo, line 762: * positions are correct?

resatomo (approved):

PR #513, opened 2024-02-03
Eye target calibration

resatomo:
Computes regression coefficients using eye positions and target positions


aopy/preproc/base.py
```diff
@@ -732,3 +732,58 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+    """
+    Extracts eye data and calibrates, aligning eye data and calculating the least square fitting coefficients 
+    between eye positions and target positions
+    
+    Args:
+        eye_data (nt, 2 or 4): eye data in time (optionally for both left and right eyes)
+        eye_samplerate (float): sampling rate of the eye data
+        event_times (nevent): times at which events occur
+        event_codes (nevent): codes for each event
+        target_pos (nevent, 3): target positions (x,y,z) for each event
```
leoscholl, line 747: not sure how we will get these for every event. i suggest this to be a (ntarget, 3) array indexed by target index -- you can then find the target location at each trial by indexing into target_pos with the event code for cursor entering target (minus the event code for cursor enter center target)

aopy/preproc/base.py
```diff
@@ -732,3 +732,58 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+        align_events (list, optional): list of event codes to use for alignment. By default, align to
+            when the cursor enters 8 peripheral targets
+        trial_end_events (list, optional): list of end events to use for alignment. By default trial end is code 239
```
leoscholl, line 750: i recommend using trial_separate instead of get_trial_segments for this alignment, that way you can more easily exclude trials. see the recently updated calc_eye_calibration() function above

aopy/preproc/base.py
```diff
@@ -732,3 +732,58 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+        offset (float, optional): time (in seconds) to offset from the given events to correct for a delay in eye movements
```
leoscholl, line 751: offset from what? does positive offset yield earlier or later eye data?

aopy/preproc/base.py
```diff
@@ -732,3 +732,58 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+        raise ValueError("Not enough trials to calibrate")
+    align_times = trial_times[:,0] + offset
+    start_time  = (align_times * eye_samplerate).astype(int)
+    end_time =  ((align_times + duration) * eye_samplerate).astype(int)
+    if duration > 0:
+        eye_data_aligned = np.array([np.nanmean(eye_data[start:end,:],axis=0) for start, end in zip(start_time,end_time)])
+    else:
+        eye_data_aligned = eye_data[start_time,:]
+        valid_indexes = ~np.isnan(start_time)
+        start_time = start_time[valid_indexes].astype(int)
```
leoscholl, line 773: i'm not sure why start_time would be NaN? can you make a comment to explain? also should you move these two lines earlier so it applies to all the data and not just if the duration == 0?

tests/test_preproc.py
```diff
@@ -603,6 +603,21 @@ def test_calc_eye_calibration(self):
 
         np.testing.assert_allclose(expected_coeff, coeff)
         np.testing.assert_allclose(expected_corr_coeff, corr_coeff)
+        
+    def test_calc_eye_target_calibration(self):
+        eye_data = np.array([[0, 1, 2, 3, 4, 5, 6, 7], [3, 4, 5 ,6 ,7, 8, 9, 10]]).T
+        eye_samplerate = 1
+        event_times = [2, 3, 4, 5]
+        event_codes = [0, 1, 0, 1]
+        target_pos_test = np.array([[2,5,0],[4,7,0]])
```
leoscholl, line 612: in the real data you won't have a target position for every event. i suggest adding some irrelevant event codes to your test. e.g. event_times = [2, 3, 4, 5, 6, 7, 8] and event_codes = [0, 2, 1, 3, 0, 5, 1]

aopy/preproc/base.py
resatomo: It didn't make sense to me, either. I used the same code as calc_eye_calibration.

resatomo: I removed these parts

aopy/preproc/base.py
```diff
@@ -729,3 +729,63 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+    """
+    Extracts eye data and calibrates, aligning eye data and calculating the least square fitting coefficients 
+    between eye positions and target positions
+    
+    Args:
+        eye_data (nt, 2 or 4): eye data in time (optionally for both left and right eyes)
+        eye_samplerate (float): sampling rate of the eye data
+        event_times (nevent): times at which events occur
+        event_codes (nevent): codes for each event
+        target_pos (ntarget, 3): N target positions (x,y,z)
```
leoscholl, line 744: Can you include an explanation that they should be ordered by their target idx

aopy/preproc/base.py
```diff
@@ -729,3 +729,63 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+        tuple: tuple containing:
+            | **coefficients (neyech, 2):** coefficients [slope, intercept] for each eye channel
+            | **correlation_coeff (neyech):** correlation coefficients for each eye channel
+    """
+
+    # Get the corresponding cursor and eye data
+    trial_events, trial_times= trial_separate(event_codes, event_times, align_events, n_events=2)
+    if len(trial_times) == 0:
+        raise ValueError("Not enough trials to calibrate")
+    exclude_trials = np.array([(t[-1] in penalty_events) for t in trial_events], dtype='bool')
```
leoscholl, line 764: This might be more flexible if we use “np.any(np.isin(t, penalty_events))” instead of “t[-1] in penalty_events”

aopy/preproc/base.py
```diff
@@ -729,3 +729,63 @@ def calc_eye_calibration(cursor_data, cursor_samplerate, eye_data, eye_samplerat
...
+    align_times = trial_times[~exclude_trials,0] + offset
+    
+    start_time  = (align_times * eye_samplerate).astype(int)
+    end_time =  ((align_times + duration) * eye_samplerate).astype(int)
```
leoscholl, line 768: I don’t think this will work with duration=0, you might need to add 1 to end_time. Would be nice to include a test

leoscholl (approved):
Looks good made some minor suggestions

PR #514, opened 2024-02-05
add kwargs to load_chmap

leoscholl:


aopy/data/base.py
```diff
@@ -744,7 +745,7 @@ def load_chmap(drive_type='ECoG244', acq_ch_subset=None, theta=0):
         layout = pd.read_excel(f)
     if acq_ch_subset is not None:
         acq_ch_subset = np.array(acq_ch_subset, dtype='int')
-    return map_acq2pos(signal_path, layout, acq_ch_subset=acq_ch_subset, theta=theta)
+    return map_acq2pos(signal_path, layout, acq_ch_subset=acq_ch_subset, theta=theta, **kwargs)
```
rcanfield2, line 748: Does map_acq2pos need to have a **kwargs input?

rcanfield2 (approved):

aopy/data/base.py
leoscholl: Not necessarily, I mainly want to have access to the column names so i can access row,col instead of x,y

PR #516, opened 2024-02-15
Katherine tracking task analysis

katherineperks:
Add functions to compute user transfer functions and controllers, and related helper functions


aopy/analysis/base.py
```diff
@@ -366,10 +366,34 @@ def calc_freq_domain_amplitude(data, samplerate, rms=False):
     freq_data = np.fft.fft(data, axis=0)
     length = np.shape(freq_data)[0]
     freq = np.fft.fftfreq(length, d=1./samplerate)
-    data_ampl = abs(freq_data[freq>=0,:])*2/length # compute the one-sided amplitude
+
+    # Only take non-negative frequency components
     non_negative_freq = freq[freq>=0]
```
mmadduri, line 371: do you need this variable?

aopy/analysis/base.py
```diff
@@ -366,10 +366,34 @@ def calc_freq_domain_amplitude(data, samplerate, rms=False):
...
+    non_negative_freq_data = freq_data[freq>=0,:]/complex(length,0) # normalize by length
+    non_negative_freq_data[1:,:] = non_negative_freq_data[1:,:]*2 # account for half the peak amplitude being at the negative frequency component
```
mmadduri, line 373: to be clear, it's [1: ] bcz fft gives you dc bias for it's first variable?

aopy/analysis/controllers.py
```diff
@@ -0,0 +1,204 @@
...
+        M = 1./np.ones((len(s),))
+    elif order == 1:
+        M = 1./(s)
+    elif order == 2:
+        M = 1./(s**2+s)
+    else:
+        print('Order not recognized!')
+
+    return M[exp_idx]
+
```
mmadduri, line 39: nice, this looks good.

aopy/analysis/base.py
mmadduri: also confused what happens if freqs is an odd number? or is it never an odd number?

aopy/analysis/controllers.py
```diff
@@ -0,0 +1,204 @@
...
+    for i in range(len(ref_freqs)):
+        curr_r = ref_freqs[i]
+        curr_d = dis_freqs[i]
+        # Find next trial with complementary frequency content
+        for j in range(i+1,len(ref_freqs)):
+            next_r = ref_freqs[j]
+            next_d = dis_freqs[j]
+            if (next_r != curr_r).all() and (next_d != curr_d).all() and j-i <= max_trial_distance:
+                trial_pairs.append([i,j])
+                break
```
mmadduri, line 104: why do you need the break here?

aopy/analysis/base.py
katherineperks: Which variable? `non_negative_freq`?

aopy/analysis/base.py
katherineperks: Yes, because the 0th index is the dc offset

katherineperks: Not sure I understand what you mean by if freqs is an odd number?

aopy/analysis/controllers.py
katherineperks: Good catch! I had more code underneath this in the past and forgot to remove the `break`

tests/test_analysis.py
```diff
@@ -284,6 +284,52 @@ def test_calc_rms(self):
...
+        self.assertAlmostEqual(abs(freqvalues[freqs==f1,0])[0], A1)
+        self.assertAlmostEqual(np.angle(freqvalues[freqs==f1,0], deg=True)[0], -90)
+        self.assertAlmostEqual(abs(freqvalues[freqs==f2,0])[0], A2)
+        self.assertAlmostEqual(np.angle(freqvalues[freqs==f2,0], deg=True)[0], 0)
+        ax[3].plot(freqs, abs(freqvalues), '-o', color='tab:green')
+        ax[3].set_ylabel('magnitude'); ax[3].set_xlabel('frequency')
+
+        filename = 'freq_domain_decomposition.png'
+        aopy.visualization.savefig(docs_dir, filename)
+
```
mmadduri, line 332: is this only a 1D, 1 frequency case?

katherineperks: It has 2 underlying frequencies. Not sure what you mean by 1D - I guess this would be the type of signal used in a 1D tracking task. In the case of the 2D tracking task, would you be able to use this function on the x and y components separately? Or maybe there should be another function that breaks a 2D signal into its x and y components and then calls `calc_freq_domain_values()` on each component

aopy/analysis/base.py
mmadduri: yeah non_negative_freq

aopy/analysis/controllers.py
mmadduri: okay!

aopy/analysis/controllers.py
```diff
@@ -0,0 +1,204 @@
...
+        dis_ind_b = np.isin(exp_freqs, dis_freqs[trial_b])
+
+        Tur[pair_id, ref_ind_a] = trial_Tur[trial_a, ref_ind_a]
+        Tur[pair_id, ref_ind_b] = trial_Tur[trial_b, ref_ind_b]
+        Tud[pair_id, dis_ind_a] = trial_Tud[trial_a, dis_ind_a]
+        Tud[pair_id, dis_ind_b] = trial_Tud[trial_b, dis_ind_b]
+
+        B[pair_id,:] = np.divide( -Tud[pair_id], np.multiply( M, np.ones(Tud[pair_id].shape,dtype=complex)+Tud[pair_id] ) )
+        F[pair_id,:] = np.multiply( Tur[pair_id], (1+0j)+np.multiply( B[pair_id,:], M ) ) - B[pair_id,:]
+
```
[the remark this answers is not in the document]
mmadduri: This is how it's written in Burden Lab code:

   B = - (1 /  M) * (Tud / (1+ Tud)) # 
    
    F = (1 + B * M)*Tur - B # all vectorized operations
    
    return F, B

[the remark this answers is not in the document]
mmadduri: okay, looks the same to me!

tests/test_analysis.py
mmadduri: I was asking about both -- number of frequencies and number of dims. Agree, you could do a wrapper function to separate x and y, so you're okay for now I think.

mmadduri (approved):
looks good!

aopy/analysis/base.py
mmadduri: oh if the number of frequencies is odd

aopy/analysis/controllers.py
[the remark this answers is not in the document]
mmadduri: might want to add a comment for how it's written without the np's and pair_ids

tests/test_analysis.py
katherineperks: Okay cool. Yeah my test is just a 1D signal with 2 frequencies

aopy/analysis/base.py
katherineperks: Yeah it's useful to have the frequencies (hz) corresponding to the freq domain values. Another aopy function calls this one and seems like `non_negative_freq` is used there as the x-axis in plotting spectrograms. One of my functions uses `non_negative_freq` to find the indices of the experimental freqs, since we only need the transfer functions at those frequencies.

katherineperks: Yeah it's useful to have the frequencies (hz) corresponding to the freq domain values. Another aopy function calls this one and seems like `non_negative_freq` is used there as the x-axis in plotting spectrograms. One of my functions uses `non_negative_freq` to find the indices of the experimental freqs, since we only need the transfer functions at those frequencies.

aopy/analysis/base.py
katherineperks: I don't think it matters. The function will return all non-negative freqs and their corresponding freq-domain values, including the dc offset, regardless of # of freqs. I'm multiplying all the values at positive frequencies by 2 to account for there being a symmetrical value at the negative frequency component. The freq-domain values are symmetrical around freq=0, so the dc offset value doesn't need to be corrected. Does this make sense?

aopy/analysis/controllers.py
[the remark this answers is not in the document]
katherineperks: Good idea!

aopy/analysis/controllers.py
katherineperks: Actually, I reviewed this function and wrote some tests and there was a reason I added the break here. It's related to how many pairs you're okay with a single trial contributing to (in experiments where it's not set up for the ref/dis freqs to alternate on every trial), and it's a bit of a subjective decision. Maybe we can talk about it in person?

aopy/analysis/controllers.py
hachouuw: This only calculates the M at the stimulated frequencies right? Might be a bit harder to use this function if you're doing other analysis with signals at a different frequency (e.g. non-stimulus frequencies).

tests/test_analysis.py
hachouuw, line 1: I don't get the test transfer function functions. What are they for?

katherineperks: They test calc_transfer_function() using simple examples where I know the ground truth and can make sure the function returns what I expect

aopy/analysis/controllers.py
katherineperks: I changed exp_freqs to be an optional argument so that M at all frequencies is returned by default

PR #518, opened 2024-02-27
v2 prepare bmi3d function

leoscholl:
prioritizes hdf data over ecube data for sync events


leoscholl:
maybe we also need to include preprocessing into preprocessed files and print them out upon loading

resatomo (approved):

PR #519, opened 2024-02-27
fix accllr p values

leoscholl:


aopy/analysis/accllr.py
```diff
@@ -629,7 +629,7 @@ def calc_accllr_st(data_altcond, data_nullcond, lowpass_altcond, lowpass_nullcon
 
     # Finally calculate an FDR-corrected p-value across all channels
     z = (roc_auc-0.5)/roc_se # above 0.5 (above chance)
-    p_uncorrected = norm.sf(abs(z)) # one-sided
+    p_uncorrected = norm.sf(z) # one-sided
```
rcanfield2, line 632: For one sided test do you need to define if it is the 'lower' bound or 'upper' bound?

leoscholl: i think norm.sf(z) is the same as 1 - norm.cdf(z), which is 1 - the left side of the z distribution at the value z, in other words the right side. which is what i want. but maybe i'm wrong

rcanfield2: yeah i think you're correct. If you only want that one I think you should specify which side the test is taking

rcanfield2 (approved):

PR #521, opened 2024-03-13
change calc_corr_over_distance inputs

leoscholl:
it was weird to have this function always prune the input data using acq_ch. getting rid of that argument and changing acq_data to elec_data makes it easier to input arbitrary channel data, e.g. after removing bad channels. also added a plotting function to make a correlation plot for each entry in a list.


pavi-rajes (approved):
changes look good.

PR #524, opened 2024-03-20
gaussian kernel convolution

rcanfield2:


rcanfield2:
Redoing with new pull request.

PR #525, opened 2024-03-22
Katherine tracking task analysis

katherineperks:
Revise `get_machine_dynamics()`:
1. raise error instead of just printing warning when given an unrecognized system order
2. move rounding of frequencies (because np.fft.fft sometimes leads to weird decimals on what should be whole number frequencies) to wrapper functions, not within `get_machine_dynamics()`
3. add unit tests


tests/test_analysis.py
```diff
@@ -1624,6 +1624,29 @@ def test_movement_onset_and_cursor_leave_time(self):
         self.assertTrue(np.all(cursor_leave_time == np.array([5,7])))
 
 class ControlTheoreticAnalysisTests(unittest.TestCase):
+    def test_get_machine_dynamics(self):
+        freqs = np.linspace(0,1,20)
+        exp_freqs = freqs[::2] # "even" freqs
+        print(freqs)
+        print(exp_freqs)
+        
+        M = controllers.get_machine_dynamics(freqs, 0)
```
mmadduri, line 1633: really minor -- could you add comments to what the test is? would help me figure out if the asserts make sense

mmadduri (approved):
thanks!

PR #530, opened 2024-03-29
Direction tuning etc

leoscholl:
work in progress


leoscholl:
added functions for plotting direction tuning
and added a function to get direction angles from discrete targets

aopy/analysis/tuning.py
```diff
@@ -95,6 +95,69 @@ def get_mean_fr_per_condition(data, condition_labels, return_significance=False)
     else:
         return np.array(means_d).T, np.array(stds_d).T
     
+def convert_target_to_direction(target_locations):
+    '''
```
rcanfield2, line 99: Can you add an optional argument to define where the beginning target is? I think that would make it more generalizable to other tasks, like a random target task

aopy/visualization/base.py
```diff
@@ -1425,8 +1425,143 @@ def plot_waveforms(waveforms, samplerate, plot_mean=True, ax=None):
...
+
+def plot_direction_tuning(per_direction_data, directions, show_var=True, wrap=True, ylabel='success rate', ax=None):
+    '''
+    Plot tuning curves for directional data. The mean is plotted as a solid line and the variance as 
+    a shaded region around the mean. Works with both cartesian and polar axes.
+    
+    Args:
+        mean (ndir, nch, ntrial): direction responses for each channel. If only one channel, can be (ndir, ntrial).
+        directions (ndir): unique directions in radians
+        show_var (bool, optional): if True, shows the variance around the mean. Default True.
```
rcanfield2, line 1482: Might be beyond the scope of this pull request but it would be nice for the user to be able to input their own error. For example, if they wanted to show SD instead of variance.

leoscholl: maybe someone else can do that :)

aopy/analysis/tuning.py
rcanfield2: Sorry, I don't think I was clear. The zero axis and direction are definitely important and I'm glad ty, but so is the position of the start target in x,y coordinates. Right now it seems like the start target is always assumed to be at the center of the screen.

For example, if the target is at (0,5) (directly up from the center of the screen ) the angle will be 270 if the start target is at the center of the screen. If the start target is at (5,0) (directly right from the center of the screen) the angle will be 235.

leoscholl: could we call that the "origin"? default (0,0)

rcanfield2: yup!

leoscholl: what should be the angle if the target is at the origin? np.nan?

rcanfield2: yeah probably, and then in the docstring say that nans will occur if the target is at the origin input by the user. (Maybe a warning as well?)

aopy/analysis/tuning.py
```diff
@@ -95,6 +96,79 @@ def get_mean_fr_per_condition(data, condition_labels, return_significance=False)
...
         return np.array(means_d).T, np.array(stds_d).T
     
+def convert_target_to_direction(target_locations, origin=[0,0], zero_axis=[0,1], clockwise=True):
+    '''
+    Converts target index to target direction in radians. NaNs are returned for targets at the origin.
+
+    Args:
+        target_locations (ntargets, 2): array of unique target (x, y) locations
+        origin (2-tuple): (x,y) coordinate of the center of all targets defining the polar plane. Default is [0,0].
+        zero_axis (2-tuple): (x,y) coordinate of the axis representing zero degrees. Default is [0,1] (up).
```
rcanfield2, line 106: Not a big deal but the optional argument is a list, not a tuple.

rcanfield2 (approved):

PR #531, opened 2024-04-01
Katherine tracking task behavior

katherineperks:
Add functions to calculate behavioral metrics relevant for tracking task: `calc_tracking_error()` and `calc_tracking_in_time()`


aopy/analysis/behavior.py
```diff
@@ -337,4 +337,48 @@ def get_cursor_leave_time(cursor_traj, samplerate, target_radius):
...
+        user_traj (nt,ndim): user trajectory over a trial segment
+        target_traj (nt,ndim): target trajectory over a trial segment
+
+    Returns:
+        tracking_error (ndim,): array of floats corresponding to the tracking error in each dimension
+    '''
+    assert len(user_traj) == len(target_traj), "User and target trajectories must be the same length!"
+    return np.mean((user_traj - target_traj)**2, axis=0) # compute mean over time axis
+
+def calc_tracking_in_time(event_codes, event_times, proportion=False):
```
leoscholl, line 359: why do you use the task events instead of the user and target trajectories? what if you want to only analyze a small chunk of time, for example a single trial, would this be possible?

katherineperks: I think it's simpler to use events because there are events for specifically when the cursor enters and leaves the target. If I use trajectories, I need to account for the target and cursor radii. I wrote this function intending it to be use on a single trial basis, like in my test function!

leoscholl: ok makes sense!

leoscholl (approved):
looks good!

PR #533, opened 2024-04-02
add summarize_entries

leoscholl:
a utility to make a dataframe from task entries. maybe useful for summarizing across days


docs/source/_images/db_summarize_sessions.png
katherineperks, line 1: Do you expect the duration of resting data experiments to always be 0min?

katherineperks (approved):
Looks good! Let's make a note that database tests shouldn't be run on the servers

PR #536, opened 2024-04-25
function to do permutation test added in analysis

pavi-rajes:
Added a function to perform permutation test. Current code takes any statistics function as input as long as the first variable returned by the statistics function is the statistic.

### DIFF
--- a/tests/test_analysis.py
+++ b/tests/test_analysis.py
@@ -1852,7 +1852,7 @@ class stats(unittest.TestCase):
 
         # Perform permutation test using ttest_ind from scipy.stats directly
         p_value = aopy.analysis.permutation_test(sample1, sample2, ttest_ind, num_permutations)
-        self.assertAlmostEqual(p_value, 0.998)
+        self.assertAlmostEqual(p_value, 0.997)
 
 if __name__ == "__main__":
 


aopy/analysis/base.py
```diff
@@ -1667,3 +1667,46 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+
+
+'''
+Statistics 
+'''
+def permutation_test(sample1, sample2, statistic, num_permutations):
+    """
+    Conducts a permutation test to compute the p-value for the t-test statistic.
+
+    Parameters:
```
leoscholl, line 1679: Args:

aopy/analysis/base.py
```diff
@@ -1667,3 +1667,46 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+        sample1 (array-like): distribution 1.
+        sample2 (array-like): distribution 2.
+        statistic (function): The function to compute the test statistic.
```
leoscholl, line 1682: maybe call this statistic_fn to be more consistent with other function arguments in aopy

aopy/analysis/base.py
```diff
@@ -1667,3 +1667,46 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+                              This function should take two arrays as input
+                              (representing the samples) and return the test statistic.
+        num_permutations (int): The number of permutations to perform.
+
+    Returns:
+        p_value (float): The p-value for the permutation test.
```
leoscholl, line 1688: can you be more descriptive about the p-value, is it the probability of the null distribution being above the original sample statistic?

tests/test_analysis.py
```diff
@@ -1839,6 +1839,21 @@ def test_pair_trials_by_frequency(self):
...
+
+        sample1 = np.array([1, 2, 3, 4, 5])
+        sample2 = np.array([6, 7, 8, 9, 10])
+
+        # Define the number of permutations
+        num_permutations = 1000
+
+        # Perform permutation test using ttest_ind from scipy.stats directly
+        p_value = aopy.analysis.permutation_test(sample1, sample2, ttest_ind, num_permutations)
+        self.assertAlmostEqual(p_value, 0.997)
```
leoscholl, line 1855: i'm a bit confused by this, i expected a small p-value if the samples are from different distributions

leoscholl:
thanks for adding this! seems like a more intuitive and simplified version of the scipy function https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.permutation_test.html

maybe you could point to that function in the docstring in case folks need more control over the permutation test

aopy/analysis/base.py
```diff
@@ -1667,3 +1667,46 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+
+        # Compute the t-test statistic for the permuted samples
+        perm_stat, _ = statistic(perm_sample1, perm_sample2)
+        perm_stats[i] = perm_stat
+
+    # Calculate the p-value as the proportion of permutation test statistics
+    # that are more extreme than the observed statistic
+    p_value = np.sum(perm_stats > obs_stat) / num_permutations
+
+    return p_value
```
resatomo, line 1712: I also want null distribution and observed statistic.

aopy/analysis/base.py
```diff
@@ -1667,3 +1667,46 @@ def calc_corr2_map(data1, data2, knlsz=15, align_maps=False):
...
+        perm_sample1 = combined[:len(sample1)]
+        perm_sample2 = combined[len(sample1):]
+
+        # Compute the t-test statistic for the permuted samples
+        perm_stat, _ = statistic(perm_sample1, perm_sample2)
+        perm_stats[i] = perm_stat
+
+    # Calculate the p-value as the proportion of permutation test statistics
+    # that are more extreme than the observed statistic
+    p_value = np.sum(perm_stats > obs_stat) / num_permutations
```
resatomo, line 1710: Is it possible to add two-sided test as well?

PR #537, opened 2024-04-29
saving neural data segments example

leoscholl:
added an example


rcanfield2 (approved):
I can't see the ipynb

PR #538, opened 2024-05-04
Changed so that target index could be aligned in the ascending order

resatomo:
This function chooses trials so that the number of trials per target can be the same across targets. But the order of target index was completely random. This sometimes caused problems when I did decoding analysis. I changed the code so that target index could be aligned like [1,2,3,4,5,6,7,8,1,2,3,4,5,...], not completely random order.

### DIFF
--- a/aopy/postproc/base.py
+++ b/aopy/postproc/base.py
@@ -169,15 +169,15 @@ def get_minimum_trials_per_target(target_idx, cond_mask=None):
     
     return min_trial
 
-def get_conditioned_trials_per_target(target_idx, min_trial, cond_mask=None, replacement=False, seed=None):
+def get_conditioned_trials_per_target(target_idx, trials_per_cond, cond_mask=None, replacement=False, seed=None):
     '''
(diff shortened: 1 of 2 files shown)


aopy/postproc/base.py
```diff
@@ -185,22 +186,24 @@ def get_conditioned_trials_per_target(target_idx, min_trial, cond_mask=None, rep
...
-    for itarget in np.unique(target_idx):
+    for idx, itarget in enumerate(np.unique(target_idx)):
         if cond_mask is None:
             trial_mask_targ = np.where(target_idx == itarget)[0]
         else:
             trial_mask_targ = np.where(cond_mask * (target_idx == itarget))[0] 
                    
         if trial_mask_targ.size:
-            tmp = np.concatenate([tmp, np.random.choice(trial_mask_targ, min_trial, replace=replacement)])
+            if idx == 0:
```
rcanfield2, line 200: Now the input arg "min_trial" is not an accurate description, I think you should rename it to something like "trials_per_cond"

rcanfield2 (approved):

PR #540, opened 2024-05-06
update kinematics upsampling

leoscholl:
add kwargs


docs/source/_images/downsample.png
katherineperks, line 1: You can avoid the phase shift in downsampled data when using get_interp_kinematics by setting upsamplerate=samplerate, so that it doesn't do any up/down sampling, only interpolation at the same samplerate.

katherineperks (approved):
Looks good. We will revisit later: 1) phase-shift in the downsample function, 2) get_interp_kinematics doesn't up/downsample data, only interpolates, for eye data.

PR #544, opened 2024-05-07
Get a confidence interval and compute overlap ratio between 2 confidence intervals

resatomo:
Computes a confidence interval from cdf in data samples, not a confidence interval in the mean of samples.
Computes a overlap ratio (0 to 1) between 2 confidence intervals


pavi-rajes (approved):
functions and tests look good.

PR #551, opened 2024-06-10
Extract features

leoscholl:
- added a more generic function to load interpolated task data - `get_task_data`
- get online extracted features from a single recording with `get_extracted_features`
- get online decoded states from a single recording with `get_decoded_states`
- extract offline features from a single recording with `extract_lfp_features`
- added a function to tabulate task data `tabulate_task_data`
- tabulate extracted features across recordings with `tabulate_feature_data`
- tabulate decoded states across recordings with `tabulate_state_data`
- tabulate offline extracted features from across recordings with `tabulate_lfp_features`

Tried to ensure the offline features are close to online features as possible. Using broadband data is slightly better than lfp data for reconstructing offline states. 
download-19

Latency was set at default to 20 ms after titrating across a single recording.
download-16

Also careful that the tabulated features and states are the same as the ones from a single recording, but note they are not identical for a KF.


aopy/data/bmi3d.py
```diff
@@ -1004,7 +1117,176 @@ def get_ts_data_segment(preproc_dir, subject, te_id, date, start_time, end_time,
 
     return data, samplerate
 
+@lru_cache(maxsize=1)
+def extract_lfp_features(preproc_dir, subject, te_id, date, decoder, samplerate=None, channels=None,
```
rcanfield2, line 1121: It isn't clear that the decoding is happening in this function based on the name. I think you should either just rename to something like "tabulate_lfp_features_and_decode" or pull the decoding out of this function and make a wrapper.

rcanfield2: Actually, since this returns either features or states I think they should be separate functions.

aopy/data/bmi3d.py
```diff
@@ -1716,6 +1969,284 @@ def tabulate_stim_data(preproc_dir, subjects, ids, dates, metadata=['stimulation
...
+    for s, t, d, ts, te in zip(subjects, te_ids, dates, start_times, end_times):
+        task_data, samplerate = get_task_data(
+            preproc_dir, s, t, d, datatype, samplerate=samplerate, step=step,
+            preproc=preproc, **kwargs) # cached for each unique recording
+        segments.append(get_data_segment(task_data, ts, te, samplerate))
+
+    segments = np.array(segments, dtype='object')
+    return segments, samplerate
+
+def tabulate_lfp_features(preproc_dir, subjects, te_ids, dates, start_times, end_times,
```
rcanfield2, line 2038: Note that the same decoder is used for each te_id even though that may not have been what happened in practice. Also, it would be nice to be able to input multiple different decoders; maybe a list the same length as te_ids that has decoders for each te_id

rcanfield2, line 2038: Also, it'd be good to note that the decoder isn't reset to 0 between segments/te_ids so decoding the states from features will likely have problems for a KF

rcanfield2 (commented):
The more I think about it the more I think that the 'decode' functionality should just be removed from this pull request.

leoscholl:
i removed the 'decode' parameter (and kept a secret hidden function _extract_lfp_features() with 'decode') and updated the tabulate functions to input a list of decoders as well as a single decoder.

rcanfield2 (approved):

PR #553, opened 2024-06-18
Pavi centerout postproc

pavi-rajes:
This branch is work in progress. I added a centerout.py file in postproc to include functions related to getting the mapping used and perform coordinate transformations.
Need to write test cases for all functions in centerout.py. I have included test cases for coordinate transformation functions. those two can be merged.


aopy/postproc/centerout.py
```diff
@@ -0,0 +1,237 @@
...
+    T = np.array([
+        [0, 0, 1],
+        [0, 1, 0],
+        [1, 0, 0]
+    ])
+
+    # Perform the matrix multiplication
+    h_coords = o_coords.dot(T)
+
+    return h_coords
```
leoscholl, line 211: shouldn't this also include the offset and scale portion of the full `_transform_coords()`?

leoscholl (commented):
get_mapping and transform_coords are implemented in #594
- it would be great to make a new PR with the images you made describing the various coordinate systems. we defined "raw" coordinate system to mean the coordinate system used by inputs like optitrack, the "bmi3d" coordinate system with y/z flipped, the "screen" coordinate system that is xyz=right/up/into the screen, and the "world" coordinate system xyz=right/up/forwards for the user.
- it would also be great to add get_taskspace_and_nullspace and decompose_hand_movements to aopy

PR #561, opened 2024-06-27
plot_decoders

leoscholl:
WIP


leoscholl:
added a function `plot_decoder_weights`
- should work with both kf and lindecoders
- plots both position and velocity vectors
- can specify a colormap

aopy/visualization/bmi3d.py
```diff
@@ -0,0 +1,111 @@
...
+    im.set_clim(-max_weight, max_weight)
+    
+def plot_readout_map(decoder, readouts, drive_type='ECoG244', cmap='YlGnBu', ax=None):
+    """
+    Plot the spatial location of readouts.
+
+    Args:
+        decoder (riglib.bmi.Decoder): The decoder object from BMI3D.
+        readouts (list): The readout channels.
+        drive_type (str, optional): The type of drive. Defaults to 'ECoG244'.
```
rcanfield2, line 41: It only handles ECoG244 now

aopy/visualization/bmi3d.py
```diff
@@ -0,0 +1,111 @@
...
+    max_len = np.max(np.linalg.norm(matrix[[x_idx,y_idx],:], axis=0))
+    if max_len == 0:
+        max_len = 1.
+    nch = matrix.shape[1]
+    ax.quiver(np.zeros(nch), np.zeros(nch), matrix[x_idx, :], matrix[y_idx, :], width=.01, 
+                 color=colors, alpha=0.5, angles='xy', scale_units='xy', scale=1.)
+    ax.set(xlabel='Px weight', ylabel='Py weight', 
+           xlim=(-max_len,max_len), ylim=(-max_len,max_len))
+
+def plot_decoder_weights(decoder, drive_type='ECoG244', cmap='YlGnBu'):
```
rcanfield2, line 82: this function name is a little ambiguous. I think it should be something like "plot_decoder_weights_wrapper" to indicate that its doing multiple plots.

leoscholl: how about `plot_decoder_summary`

PR #572, opened 2024-09-13
synchronize time series data

resatomo:
synchronize time series data by padding or cropping

### DIFF
--- a/aopy/preproc/neuropixel.py
+++ b/aopy/preproc/neuropixel.py
@@ -223,3 +223,36 @@ def concat_neuropixel_within_day(np_datadir, kilosort_dir, subject, date, ch_con
         print('No data to concatenate')
 
     return savedir_names
+
+def sync_ts_data(data, sync_timestamp):
+    '''
+    Synchronize time series data by padding nan or cropping datapoints based on synchronized timestamps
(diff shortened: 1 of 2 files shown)


aopy/preproc/neuropixel.py
```diff
@@ -222,4 +222,37 @@ def concat_neuropixel_within_day(np_datadir, kilosort_dir, subject, date, ch_con
...
+        padding_datapoints = tmp.shape[0]
+        pad_to_data = np.zeros((padding_datapoints,nch))*np.nan
+        sync_data = np.concatenate([pad_to_data, data],axis=0)
+        
+    # When the initial timestamp is less than 0, crop the head of data so that data could begin at time 0
+    else:
+        not_crop_datapoints = sync_timestamp >= 0
+        sync_data = data[not_crop_datapoints,:]
+        
+    return np.squeeze(sync_data)
```
rcanfield2, line 258: Do you need to return updated timestamps if you're padding nans?

rcanfield2 (approved):

aopy/preproc/neuropixel.py
rcanfield2: It would probably be best to return them for future use even if they aren't used.

PR #575, opened 2024-09-26
Added port_number

resatomo:
Added port_number for multiple recordings in get_preprocessed_filename and load_preproc_lfp_data


leoscholl:
is it possible to concatenate data from multiple ports into one preprocessed file?

PR #577, opened 2024-10-29
add metadata columns function

leoscholl:
maybe useful utility to add one or more columns to a dataframe with metadata from task entries from the database. i envision it being used after you create a df using tabulate_behavior_data functions but want extra info about the experiment rotations, e.g.


katherineperks (commented):
Code looks good. But could you add more info in the docstring about the potential functions that can be called on a task entry? Is there somewhere in bmi3d that lists all these attributes of a session? I think it could be confusing since the options for metadata to include here are different than that in tabulate functions

katherineperks:
Never mind my last sentence, I understand how you can get metadata about the task params! Is there a place where all these attributes of an entry are listed?
image

katherineperks (approved):

PR #579, opened 2024-11-02
Multiply matrix by huge data

resatomo:
Multiply a matrix by long data in each batch to save memory.

### DIFF
--- a/aopy/utils/base.py
+++ b/aopy/utils/base.py
@@ -767,7 +767,7 @@ def convert_port_number(port_number, datatype='ap'):
     
     return probe_dir
 
-def multiply_mat_batch(data, mat, save_path, sample_rate, scale = 1, batch_length = 2, dtype='int16'):
+def multiply_mat_batch(data, mat, save_path, scale = 1, max_memory_gb = 1., dtype='int16'):
     '''
     Multiply a matrix to data in each batch to save memory. The result is saved in save_path.
     This function can be used to multiply an inverse matrix by spike band time series.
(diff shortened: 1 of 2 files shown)


aopy/utils/base.py
```diff
@@ -765,4 +765,39 @@ def convert_port_number(port_number, datatype='ap'):
...
+    Multiply a matrix to data in each batch to save memory. The result is saved in save_path.
+    This function can be used to multiply an inverse matrix by spike band time series.
+    
+    Args:
+        data (nt, nch): neural data. This should be a memory mapping array.
+        mat (anysize, nch): matrix to multiply by data
+        save_path (str): file path to save destriped lfp data
+        sample_rate (float): sampling rate in Hz
+        scale (float, optional): Scaling factor to multiply by data. 1/200 is necessary for whitened data in kilosort4. default is 1.
+        batch_length (int, optional): batch size in second. default is 2 seconds.
```
rcanfield2, line 781: Similar to the other pull request, I think there should at least be an option to define the batch size based on available memory.

aopy/utils/base.py
```diff
@@ -765,4 +765,39 @@ def convert_port_number(port_number, datatype='ap'):
...
+    '''
+    batch_size = int(sample_rate * batch_length)
+    n_samples, n_channels = data.shape
+    Nbatches = np.ceil(n_samples/batch_size).astype(int)
+
+    # Multiply matrix in each batch to save memory
+    x = np.memmap(save_path, dtype=dtype, mode='w+', shape=(n_samples, n_channels))
+
+    for ibatch in range(Nbatches):
+        if ibatch < Nbatches-1:
```
rcanfield2, line 796: I don't think the if/else statement is required.

aopy/utils/base.py
```diff
@@ -765,4 +765,34 @@ def convert_port_number(port_number, datatype='ap'):
...
+    
+    '''
+        
+    n_samples, n_channels = data.shape
+    batch_size = int (max_memory_gb*1e9 / (n_channels*np.dtype(dtype).itemsize))
+    Nbatches = np.ceil(n_samples/batch_size).astype(int)
+
+    # Multiply matrix in each batch to save memory
+    x = np.memmap(save_path, dtype=dtype, mode='w+', shape=(n_samples, n_channels))
+
```
rcanfield2, line 794: can you add the min batch size argument like you did in the other pull request?

resatomo: min batch size is not necessary for this function because this is just multiplication between 2 matrices.

rcanfield2 (approved):

PR #580, opened 2024-11-05
destripe lfp batch

resatomo:
Devide data into batches and destripe lfp data at each batch.

### DIFF
--- a/aopy/preproc/neuropixel.py
+++ b/aopy/preproc/neuropixel.py
@@ -242,7 +242,7 @@ def sync_ts_data_timestamps(data, sync_timestamps):
         
     return np.squeeze(sync_data), sync_timestamps
 
-def destripe_lfp_batch(lfp_data, save_path, sample_rate, bit_volts, batch_length = 180, dtype='int16'):
+def destripe_lfp_batch(lfp_data, save_path, sample_rate, bit_volts, max_memory_gb = 1., dtype='int16'):
     
     '''
     Destripe LFP data in each batch to save memory. The result is saved in save_path.
@@ -252,7 +252,7 @@ def destripe_lfp_batch(lfp_data, save_path, sample_rate, bit_volts, batch_length
         save_path (str): file path to save destriped lfp data
         sample_rate (float): sampling rate in Hz
         bit_volts (float): volt per bit
-        batch_length (int): batch size in second
+        max_memory_gb (float): memory size in GB to determine batch size. default is 1.0 GB.
     
     Returns:
         None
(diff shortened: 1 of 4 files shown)


aopy/preproc/neuropixel.py
```diff
@@ -239,4 +240,41 @@ def sync_ts_data_timestamps(data, sync_timestamps):
...
+    '''
+    Destripe LFP data in each batch to save memory. The result is saved in save_path.
+    
+    Args:
+        lfp_data (nt, nch): LFP data. This should be a memory mapping array.
+        save_path (str): file path to save destriped lfp data
+        sample_rate (float): sampling rate in Hz
+        bit_volts (float): volt per bit
+        batch_length (int): batch size in second
+
```
rcanfield2, line 256: I think it would be more useful/intuitive to set the batch length based on memory available. The input argument would be something like "memory_available" in GB, and then would calculate the batch length based on the number of channels and data type. This way the function would always run the fastest allowable based on the memory available. 

What do you think? If we do this, would we need a minimum batch length for the destriping function to work?

aopy/preproc/neuropixel.py
```diff
@@ -239,4 +240,41 @@ def sync_ts_data_timestamps(data, sync_timestamps):
...
+    batch_size = int(sample_rate * batch_length)
+    Nbatches = np.ceil(n_samples/batch_size).astype(int)
+
+    for ibatch in range(Nbatches):
+        if ibatch < Nbatches-1:
+            tmp = destripe_lfp((lfp_data[ibatch*batch_size:(ibatch+1)*batch_size, :]*bit_volts).T/1e6, sample_rate)*1e6
+            lfp_destriped[ibatch*batch_size:(ibatch+1)*batch_size, :] = (tmp/bit_volts).T.astype(dtype)
+        else:
+            tmp = destripe_lfp((lfp_data[ibatch*batch_size:, :]*bit_volts).T/1e6, sample_rate)*1e6
+            lfp_destriped[ibatch*batch_size:, :] = (tmp/bit_volts).T.astype(dtype)
```
rcanfield2, line 278: I don't think this if/else statement is necessary. If you index into an array with a range that ends larger than the size of the array it should still work.

tests/test_preproc.py
```diff
@@ -1721,7 +1722,21 @@ def test_sync_ts_data_timestamps(self):
...
+
+        data = np.random.randint(0,50, (10000,384))
+        sample_rate = 1000
+        bit_volts=0.1
+
+        destripe_lfp_batch(data, save_path, sample_rate, bit_volts, batch_length = 180, dtype='int16')
+        destriped_data = np.memmap(save_path, shape=data.shape, dtype='int16')
+
+        print(np.all(data.shape == destriped_data.shape))
+        del destriped_data # delete memmap file to completely delete the file later
```
rcanfield2, line 1737: Tests should use an 'assert' so that the test fails if the condition isn't true.

tests/test_preproc.py
```diff
@@ -1721,7 +1722,21 @@ def test_sync_ts_data_timestamps(self):
         dt = sync_timestamp[1] - sync_timestamp[0]
         num_padding = np.arange(sync_timestamp[0]-dt,0,-dt).shape[0]
         self.assertTrue(sync_data.shape[0], num_padding + sync_timestamp.shape[0])  
-              
+    
+    def test_destripe_lfp_batch(self):
```
rcanfield2, line 1726: I think we also need a test that the function is grabbing the batches correctly. Maybe just make sure that each value in the destriped array is different from the value in the original array?

resatomo: I checked how much the destriped array was different from the original array. I found values of 1% elements changed from the original array. Almost all of elements in the destriped array was the same as the original elements. I'm not sure what test is appropriate for this function.

aopy/preproc/neuropixel.py
```diff
@@ -3,6 +3,7 @@
 from ..data import load_neuropixel_data, load_ks_output, get_channel_bank_name
 import os
 import glob
+from ibldsp.voltage import destripe_lfp
```
rcanfield2, line 6: Do we need to add idldsp to setup.py? (and mock imports in the docs)

aopy/preproc/neuropixel.py
```diff
@@ -239,4 +240,37 @@ def sync_ts_data_timestamps(data, sync_timestamps):
...
+
+    # Load data and metadata
+    n_samples, n_channels = lfp_data.shape
+
+    # Create memmap array to save destriped lfp data
+    lfp_destriped = np.memmap(save_path, dtype=dtype, mode='w+', shape=lfp_data.shape)
+
+    # Destripe lfp
+    batch_size = int (max_memory_gb*1e9 / (n_channels*np.dtype(type(bit_volts)).itemsize))
+    batch_size += 21 # ensure that batch size is more than 21 to run destrip_lfp
```
rcanfield2, line 270: Can you make the min batch size of 21 an optional argument to the function?

aopy/preproc/neuropixel.py
resatomo: I added ibllib in setup.py and conf.py in docs. Is this enough?

rcanfield2: yup I think its good!

tests/test_preproc.py
rcanfield2: Hmm I guess ideally we'd have a before and after image of neuropixel data that has been destriped. I'm not sure if that is necessary now, but could you add a link to the ibl white paper in the doc string for the function? (https://doi.org/10.6084/m9.figshare.19705522.v4)

rcanfield2 (approved):

PR #582, opened 2024-11-07
add np tabulate func and tests

rcanfield2:


rcanfield2:
not sure how the figures got in this pull request.

aopy/data/bmi3d.py
```diff
@@ -2341,7 +2426,35 @@ def tabulate_ts_segments(preproc_dir, subjects, te_ids, dates, start_times, end_
         segments.append(segment)
         
     return segments, samplerate
+
+def tabulate_spike_data_segments(preproc_dir, subjects, te_ids, dates, ports, start_times, end_times, bin_width=0.01):
+    '''
+        Grab nonrectangular timeseries data from trials across arbitrary preprocessed files.
```
leoscholl, line 2432: indent is wrong

aopy/data/bmi3d.py
```diff
@@ -1117,6 +1117,91 @@ def get_ts_data_segment(preproc_dir, subject, te_id, date, start_time, end_time,
 
     return data, samplerate
 
+def get_spike_data_segment(preproc_dir, subject, te_id, date, port, start_time, end_time, bin_width=.01):
```
leoscholl, line 1120: for future compatability with tomo's PR, can you change port to drive and make it an optional argument? you could just have drive=1 for now by default

aopy/data/bmi3d.py
```diff
@@ -1117,6 +1117,91 @@ def get_ts_data_segment(preproc_dir, subject, te_id, date, start_time, end_time,
...
+    Returns:
+        tuple: A tuple containing:
+            - spike_segment (dict): A dictionary where keys are unit labels and values are arrays of spike times (or binned spike counts) for that unit.
+            - bins (numpy.ndarray or None): An array of bin edges if binning was applied, otherwise `None`.
+    
+    '''
+
+    # Load data
+    filename_mc = get_preprocessed_filename(subject, te_id, date, 'spike')
+    spike_data = load_hdf_group(os.path.join(preproc_dir, subject), filename_mc, f'drive{port}/spikes')
```
leoscholl, line 1143: i think you want to set cached=True here otherwise you're throwing away the spike data and reloading it each time for the same recording but different start and end times.

leoscholl: with continuous data we load just the relevant section of data from hdf but i don't think that approach works for spike times

aopy/data/bmi3d.py
```diff
@@ -1117,6 +1117,91 @@ def get_ts_data_segment(preproc_dir, subject, te_id, date, start_time, end_time,
...
+            - spike_aligned (numpy.ndarray): A 3D array of aligned spike data with shape (ntime, nunits, ntrials), where:
+                - ntime is the number of time bins between `time_before` and `time_after` around each trigger.
+                - nch is the number of units.
+                - ntrials is the number of trials (trigger events).
+            - unit_labels (list of str): A list of unit labels corresponding to the 'nunits' dimension in the aligned spike data.
+            - bins (numpy.ndarray): The time bin centers relative to the trigger times.
+    """
+    # Load data
+    filename_mc = get_preprocessed_filename(subject, te_id, date, 'spike')
+    spike_data = load_hdf_group(os.path.join(preproc_dir, subject), filename_mc, f'drive{port}/spikes')
```
leoscholl, line 1185: add cached=True argument here too

aopy/data/bmi3d.py
```diff
@@ -1117,6 +1117,91 @@ def get_ts_data_segment(preproc_dir, subject, te_id, date, start_time, end_time,
...
+    ntime = int(np.round((time_after+time_before)/bin_width))
+    nch = len(spike_data)
+    ntrials = len(trigger_times)
+    
+    # Parse segment and bin spikes if necessary.
+    unit_labels = list(spike_data.keys())
+    spike_aligned = np.zeros((ntime, nch, ntrials))*np.nan #(ntime, nch, ntrials)
+    for iunit, unit_label in enumerate(unit_labels):
+        binned_spikes, _ = precondition.bin_spike_times(spike_data[unit_label], 0, trigger_times[-1]+time_after, bin_width)
+        spike_aligned[:,iunit,:] = np.squeeze(preproc.base.trial_align_data(binned_spikes, trigger_times-(bin_width/2), time_before, time_after, samplerate)) # Squeeze to remove singleton dimension from only doing one unit at a time. Adjust trigger times to center on the previously binned spikes.
```
leoscholl, line 1199: can you explain why you chose to do binning and then aligning rather than align first and then bin?

leoscholl: also what's the utility of this function that operates on a single file rather than a version that can take multiple recordings as input similar to tabulate_spike_data_segments below?

rcanfield2: I copied the same structure that I saw used for lfp data. "get_lfp_aligned" also only operates on one function. I could make a wrapper for this but I think there are a few things that are unclear now:

- what output is best. A concatenated array? A list of arrays? etc. 
- How would metadata be tracked (unit labels, etc)? Would we need a whole dictionary?

For these reasons, I've decided to wait until more people are using the data to see what the best path forward is.

leoscholl (approved):
looks good, make sure you have the latest matplotlib installed, maybe you are regenerating some figures with an older version. if you already have the newest version then maybe you're updating some older figures!

aopy/data/bmi3d.py
leoscholl: ok makes sense!

rcanfield2:
I have matplotlib 3.9.2, which seems to be the latest according to google? Lets see what happens

PR #583, opened 2024-11-07
Make average neuropixel lfp array + max coherence & lag array generation

espeare:
Added a function to aopy.preproc.neuropixel called make_average_neuropixel_array as well as a function to aopy.analysis.base called calc_max_coh_and_lags as well as test cases to their corresponding test files


aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+    
+    Parameters:
+    - lags: List of time lags (in seconds) to apply to the channels in lfp_array1.
+    - lfp_array1: First LFP array (e.g., motor cortex data) with shape (n_channels, n_timepoints).
+    - lfp_array2: Second LFP array (e.g., pre-motor cortex data) with shape (n_channels, n_timepoints).
+    - frequency_range: Frequency range of interest (e.g., [0, 4] Hz for delta band).
+
+    Returns:
+    - max_coherence_matrix: Matrix of maximum coherence values for each channel pair.
+    - lag_of_max_coherence_matrix: Matrix of corresponding lags (in ms) for max coherence values.
```
rcanfield2, line 1754: Please make the docstring google format or else the readthedocs won't work. (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+    """
+    # Taper parameters
+    NW, BW, step, fk, fs = 0.5, 4, 0.005, 200, 2500
```
rcanfield2, line 1757: Can you make these input arguments to the function?

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+
+def calc_max_coh_and_lags(lags, lfp_array1, lfp_array2, frequency_range):
+    """
+    Calculate the maximum coherence and corresponding lags between pairs of channels
+    in two local field potential (LFP) arrays within a specified frequency range.
+    
+    Parameters:
+    - lags: List of time lags (in seconds) to apply to the channels in lfp_array1.
+    - lfp_array1: First LFP array (e.g., motor cortex data) with shape (n_channels, n_timepoints).
+    - lfp_array2: Second LFP array (e.g., pre-motor cortex data) with shape (n_channels, n_timepoints).
```
rcanfield2, line 1749: Lab convention is to put the time axis first (even if this isn't necessarily the optimal way)

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+    print(f"Using {num_tapers} tapers, taper length {n}, half-bandwidth {half_bw}")
+
+    # Get channel counts
+    num_channels_lfp1 = lfp_array1.shape[0]
+    num_channels_lfp2 = lfp_array2.shape[0]
+
+    # Initialize matrices for max coherence values and their corresponding lags
+    max_coherence_matrix = np.zeros((num_channels_lfp1, num_channels_lfp2))
+    lag_of_max_coherence_matrix = np.zeros((num_channels_lfp1, num_channels_lfp2))
+
```
rcanfield2, line 1768: Can you initialize these arrays as np.nan values to make it obvious if the answer is 0 or if the value just didn't get filled in? You can just add "*np.nan" to the end of each of these lines.

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+    def apply_lag(signal, lag, fs):
```
rcanfield2, line 1769: Can you rename this function to something more specific? Something like "circle_shift_by_time"

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+        Positive lag shifts forward, negative lag shifts backward.
+        """
+        lag_samples = int(lag * fs)
+        return np.roll(signal, lag_samples)
+
+    def compute_coherence_for_pair(channel_lfp1, channel_lfp2):
+        """
+        Compute coherence between a specific pair of channels for all specified lags.
+        Returns coherence values for each lag at time zero.
+        """
```
rcanfield2, line 1781: Can you update the docstring as above? Also can you pull this function out of the wrapper function so that it can easily be used alone?

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+        coherence_at_zero_lag = np.empty(len(lags))
```
rcanfield2, line 1782: Similar to before, please initialize this as np.nan values. Initializing as an empty array will do weird things.

aopy/analysis/base.py
rcanfield2: Then each input value to the array can be a 1D array.

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+
+        for idx, lag in enumerate(lags):
+            # Process signal from lfp_array1 and normalize
+            signal1 = lfp_array1[channel_lfp1, :]
+            signal1 = copy.deepcopy((signal1 - np.mean(signal1)) / np.std(signal1))
```
rcanfield2, line 1787: Zscoring should be done outside of the function to make it more flexible.

aopy/analysis/base.py
```diff
@@ -1734,4 +1735,91 @@ def calc_confidence_interval_overlap(CI1, CI2):
...
+            # Find maximum coherence and corresponding lag
+            max_coherence = np.max(delta_coherence)
+            max_lag_index = np.argmax(delta_coherence)
+            max_lag = lags[max_lag_index] * 1000  # Convert lag to milliseconds
+
+            # Store max coherence and lag in matrices
+            max_coherence_matrix[ch_lfp1, ch_lfp2] = max_coherence
+            lag_of_max_coherence_matrix[ch_lfp1, ch_lfp2] = max_lag
+
+    return max_coherence_matrix, lag_of_max_coherence_matrix
```
rcanfield2, line 1825: This part can be a separate wrapper function.

rcanfield2: And can you add an optional argument called "verbose". Then, only run print statements if verbose=True

aopy/preproc/neuropixel.py
```diff
@@ -275,4 +278,86 @@ def destripe_lfp_batch(lfp_data, save_path, sample_rate, bit_volts, max_memory_g
     for ibatch in range(Nbatches):
         tmp = destripe_lfp((lfp_data[ibatch*batch_size:(ibatch+1)*batch_size, :]*bit_volts).T/1e6, sample_rate)*1e6
         lfp_destriped[ibatch*batch_size:(ibatch+1)*batch_size, :] = (tmp/bit_volts).T.astype(dtype)
-        lfp_destriped.flush()
\ No newline at end of file
+        lfp_destriped.flush()
+
+
+def make_average_neuropixel_array(te_id, date, subject, aligning_index, tbefore, tafter):
+    """
```
rcanfield2, line 285: Thanks for including this function but Tomo and I are in the middle of restructuring how LFP preprocessing is done. This will change how data is loaded and stuff so it isn't necessary to include this function in github quite yet.

tests/test_analysis.py
```diff
@@ -1905,6 +1906,49 @@ def test_calc_confidence_interval_overlap(self):
         overlap = aopy.analysis.calc_confidence_interval_overlap(CI1,CI2)
         self.assertEqual(overlap,(20-17)/(20-10))
         
+
+class TestCalcMaxCohAndLags(unittest.TestCase):
```
rcanfield2, line 1910: Can you make the name of the test class "test_{function_name_to_be_tested}"? This way it is easy to identify which set of tests belongs to which function.

tests/test_analysis.py
```diff
@@ -1905,6 +1906,49 @@ def test_calc_confidence_interval_overlap(self):
...
+        # Simulate some LFP data with random values
+        self.num_channels = 3
+        self.num_timepoints = 5000
+        self.lfp_array1 = np.random.randn(self.num_channels, self.num_timepoints)
+        self.lfp_array2 = np.random.randn(self.num_channels, self.num_timepoints)
+
+        # Mock taper parameters and the coherence calculation function
+        global precondition
+        precondition = MagicMock()
+        precondition.convert_taper_parameters = MagicMock(return_value=(256, 2, 3))  # Mock return values for tapers
```
rcanfield2, line 1925: Just import the actual function from aopy.precondition.

tests/test_analysis.py
```diff
@@ -1905,6 +1906,49 @@ def test_calc_confidence_interval_overlap(self):
...
+
+        global calc_mt_tfcoh
+        calc_mt_tfcoh = MagicMock(return_value=(
+            np.linspace(0, 200, 100),  # Mock frequency array (0-200 Hz)
+            np.linspace(-0.5, 0.5, 50),  # Mock time array (centered around 0)
+            np.random.rand(100, 50)  # Mock coherence array (freq x time)
+        ))
```
rcanfield2, line 1932: Similar to before, I'm also not sure why importing this is necessary.

tests/test_analysis.py
```diff
@@ -1905,6 +1906,49 @@ def test_calc_confidence_interval_overlap(self):
...
+        self.assertEqual(lag_of_max_coherence_matrix.shape, (self.num_channels, self.num_channels))
+
+        # Verify the values fall within expected ranges
+        self.assertTrue(np.all(max_coherence_matrix >= 0) and np.all(max_coherence_matrix <= 1), 
+                        "Coherence values should be between 0 and 1")
+
+        # Check if lags are converted to milliseconds in lag_of_max_coherence_matrix
+        self.assertTrue(np.all(np.isin(lag_of_max_coherence_matrix, np.array(self.lags) * 1000)),
+                        "Lags should be in milliseconds and match the provided lags")
+
```
rcanfield2, line 1951: These tests are good for testing that the structure of the outputs are as expected, but it doesn't require the actual values to be as expected. After pulling out the single lagged coherence function, the test will be easier to write. You can create fake signals and then circle shift them before inputting them into your function. Then you can test is the max lag is the same as the circle shift. And if it is the same signal, it should be very high. I would start something with a single hump and zeros elsewhere, but play around with it. Let me know if you have questions.

PR #584, opened 2024-11-10
np quality metric funcs and tests

rcanfield2:


aopy/postproc/neuropixel.py
```diff
@@ -0,0 +1,246 @@
...
+    for iunit, unit_lbl in enumerate(unit_labels):
+        if end_time is None:
+            relevant_spike_mask = spike_times[unit_lbl] >= start_time
+        else:
+            relevant_spike_mask = np.logical_and(spike_times[unit_lbl] >= start_time, spike_times[unit_lbl] <= end_time)
+        
+        
+        relevant_wfs = waveforms[str(unit_lbl)][relevant_spike_mask,:,:]
+        cent_wfs = relevant_wfs - np.mean(relevant_wfs, axis=1)[:,None,:] # Center each spike on each channel
+        mean_wf = np.mean(cent_wfs, axis=0)*ap_metadata['bit_volts'][0] # Mean across all spikes for each channel. becomes (ntime, nch) array
```
resatomo, line 122: Sorry. I'm now updating metadata so that variable names will be the same as the current ECoG metadata. Can you use `ap_metadata['voltsperbit'][0]`, not `ap_metadata['bit_volts'][0]`

aopy/postproc/neuropixel.py
```diff
@@ -0,0 +1,246 @@
...
+    
+    Args:
+        preproc_dir ():
+        subject (str):
+        te_id (list):
+        port (int):
+        date (date):
+        start_time (float):
+        max_time (float): [s]
+    '''
```
resatomo, line 146: Please update docstring

aopy/postproc/neuropixel.py
rcanfield2: I'll update to account for both. Should it actually be renamed `ap_metadata['microvoltsperbit']` ?

resatomo:
I think using spike_times is intuitive for other functions, not specifying preproc_dir, taskid, etc. What do you think of this?

aopy/postproc/neuropixel.py
resatomo: The current metadata in ECoG lfp uses voltsperbit. So I think we should use it.

rcanfield2:
The other functions don't actually care about the spike times, they need either the waveform amplitude or kilosort amplitude. I could have these be the inputs to the function, especially after kilosort amplitudes are added to the preprocessed data.

aopy/postproc/neuropixel.py
rcanfield2: okay, are you converting to volts? (The current scalar converts to microvolts)

resatomo: Ah, I didn't think about it. Changing scale from microvolts to volts now may be confusing?

rcanfield2: I think it probably makes sense to just convert everything to voltsperbit and update code now. @leoscholl what do you think?

leoscholl: add both!

resatomo (approved):

PR #585, opened 2024-11-12
Updated loading functions

resatomo:
Updated load_preproc_lfp_data so that it would deal with multiple recordings.
Made load_preproc_ap_data, load_preproc_spike_data, load_spike_waveforms.


leoscholl:
i think we should check which drives are present in the files and throw a warning if you're trying to load data with drive=None but there are two drives

aopy/data/base.py
```diff
@@ -159,17 +159,101 @@ def load_preproc_lfp_data(preproc_dir, subject, te_id, date, cached=True):
         subject (str): Subject name
         te_id (int): Block number of Task entry object 
         date (str): Date of recording
+        drive_number (int): drive number for multiple recordings. 1-based indexing.
 
     Returns:
-        dict: lfp data
+        ndarray: numpy array of lfp data from hdf
```
leoscholl, line 165: can you be more specific about what will be returned in the cases where only one drive or multiple drives are present

aopy/data/base.py
```diff
@@ -159,17 +159,101 @@ def load_preproc_lfp_data(preproc_dir, subject, te_id, date, cached=True):
...
+
+    Args:
+        preproc_dir (str): base directory where the files live
+        subject (str): Subject name
+        te_id (int): Block number of Task entry object 
+        date (str): Date of recording
+        drive_number (int): drive number for multiple recordings. 1-based indexing.
+
+    Returns:
+        ndarray: numpy array of ap data from hdf
```
leoscholl, line 197: same comment about return type for all these

aopy/data/base.py
```diff
@@ -159,17 +159,101 @@ def load_preproc_lfp_data(preproc_dir, subject, te_id, date, cached=True):
...
-    metadata = load_hdf_group(preproc_dir, filename, 'lfp_metadata', cached=cached)
+    
+    group_names = list_root_groups(preproc_dir, filename)
+    num_drives = len(group_names)
+        
+    if drive_number:
+        data = load_hdf_data(preproc_dir, filename, f'drive{drive_number}/lfp_data', cached=cached)
+        metadata = load_hdf_group(preproc_dir, filename, f'drive{drive_number}/lfp_metadata', cached=cached)
+    else:
+        if num_drives > 1:
```
leoscholl, line 178: shouldn't there be 2 group names for a file with lfp_data and lfp_metadata? i'm confused how this would work. it might be better to explicitly check if the group names contain 'drive1' or 'lfp_data' (or 'ap_data', etc. for other fns)

tests/test_data.py
```diff
@@ -71,6 +71,43 @@ def test_proc_eye_day(self):
...
+        cls.subject = 'test'
+        cls.date = '2024-11-12'   
+        
+    def test_load_preproc_lfp_data(self):
+        lfp_data, lfp_metadata = load_preproc_lfp_data(data_dir, self.subject, self.id, self.date, drive_number=1)
+        self.assertIsInstance(lfp_data, np.ndarray)
+        self.assertIsInstance(lfp_metadata, dict)
+        lfp_data, lfp_metadata = load_preproc_lfp_data(data_dir, self.subject, self.id, self.date, drive_number=2)        
+        self.assertIsInstance(lfp_data, np.ndarray)
+        self.assertIsInstance(lfp_metadata, dict)
```
leoscholl, line 87: make sure to add a test with drive_number=None for each of these

leoscholl: oh nevermind it's elsewhere. it's slightly confusing to break the tests for the `load_preproc_*_data` functions into two places.. could you merge them into a single class if possible?

tests/test_data.py
```diff
@@ -344,6 +381,14 @@ def test_get_hdf_dictionary(self):
         self.assertIn('group_data', result['group1'])
         print(result)
 
+    def test_list_root_groups(self):
+        testfile = 'load_hdf_contents_test.hdf'
+        group_data_dict = {'group_data': np.arange(1000)}
+        save_hdf(write_dir, testfile, data_dict=group_data_dict, data_group="/group2", append=True)
+        group_names = list_root_groups(write_dir, testfile)
+        self.assertIn('group1', group_names)
```
leoscholl, line 389: i don't think there's any guarantee that the test above will run before this test. you should make a completely new file and add both groups in this function to make sure they are there.

leoscholl:
looks good just make some more detailed comments in the docstring about the return type. it's not currently clear what the behavior will be depending on the contents of the files

resatomo:
When there are multiple drives or no drive, returns are the same. But I added additional explanation what happens when drive_number is None, but there are multiple drives.

leoscholl (approved):

PR #591, opened 2024-11-19
add coherence connectivity metrics

leoscholl:
- `get_acq_ch_near_stimulation_site()` compares spatial map channels
- `calc_connectivity_coh()` measures coherence across multiple pairs of channels
- `calc_connectivity_coh_map()` generates a map of measurements
- `plot_tf_map_grid()` plots a grid of times and frequency bands from e.g. the result from `calc_connectivity_coh_map()`


ajm2605 (approved):

aopy/analysis/connectivity.py
```diff
@@ -0,0 +1,238 @@
...
+from .. import data as aodata
+from .. import precondition
+from . import accllr
+
+def get_acq_ch_near_stimulation_site(stim_site, stim_layout='Opto32', electrode_layout='ECoG244', 
+                                 dist_thr=1, return_idx=False):
+    '''
+    Get acquisition channels near a stimulation site. Warning: this function returns channel
+    numbers, which sometimes are 1-indexed. Set the return_idx flag to True to get the channel
+    indices as well.
```
ajm2605, line 15: Is there a sop or other function that addresses ecog channel mapping in more detail? Could you potentially link/mention that here so I know where to go to understand the channel mapping?

aopy/analysis/connectivity.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Args:
+        stim_site: int, stimulation site
+        stim_layout: str, layout of stimulation sites
```
ajm2605, line 19: What are the different layout options/are they in an sop?

aopy/analysis/connectivity.py
```diff
@@ -0,0 +1,238 @@
...
+        electrode_layout: str, layout of electrodes
+        dist_thr: float or tuple, threshold for distance from stimulation site
```
ajm2605, line 21: What units is dist_thr?

aopy/analysis/connectivity.py
```diff
@@ -0,0 +1,238 @@
...
+    Args:
+        data_altcond_source (nt, n_source, ntrial): source erp data
+        data_altcond_probe (nt, n_probe, ntrial): probe erp data
+        n (float): window length in seconds
+        p (float): standardized half bandwidth in hz
+        k (int): number of DPSS tapers to use
+        fs (float): sampling rate in Hz.
+        step (float): window step size in seconds.
+        fk (float, optional): frequency range to return in Hz ([0, fk]). Default is fs/2.
+        pad (int, optional): padding factor for the FFT. This should be 1 or a multiple of 2.
```
ajm2605, line 66: What's the default?

aopy/visualization/base.py
```diff
@@ -1934,6 +1934,81 @@ def plot_tfr(values, times, freqs, cmap='plasma', logscale=False, ax=None, **kwa
...
+
+    Args:
+        freqs (nfreq,): frequency values
+        time (nt,): time values
+        tf_data (nfreq, nt, nch): time-frequency data across spatial channels
+        bands (list): list of tuples of frequency bands to plot
+        elec_pos (nch, 2): x, y position of each electrode
+        clim (tuple, optional): color limits for the plot, e.g. (0,1) for tfcoh maps. Default None
+        interp_grid (tuple, optional): (x, y) grid to interpolate the data onto. Default None
+        cmap (str, optional): colormap to use for plotting. Default 'viridis'
```
ajm2605, line 1951: Thanks for defaulting to a colorblind friendly colormap!

aopy/analysis/connectivity.py
```diff
@@ -0,0 +1,238 @@
...
+from . import accllr
+
+def get_acq_ch_near_stimulation_site(stim_site, stim_layout='Opto32', electrode_layout='ECoG244', 
+                                 dist_thr=1, return_idx=False):
+    '''
+    Get acquisition channels near a stimulation site. Warning: this function returns channel
+    numbers, which sometimes are 1-indexed. Set the return_idx flag to True to get the channel
+    indices as well.
+
+    Args:
```
rcanfield2, line 17: I don't think this is the right docstring format for args. Isn't the datatype supposed to be in parenthesis?

aopy/analysis/connectivity.py
rcanfield2: Also please explain that this can be either a single value where electrodes closer than the threshold are found, or a minimum distance can be defined with a tuple.

aopy/analysis/connectivity.py
```diff
@@ -0,0 +1,238 @@
...
+    stim_pos, stim_ch, _ = aodata.load_chmap(stim_layout)
+    stim_site_pos = stim_pos[stim_ch == stim_site]
+    if stim_site_pos.size == 0:
+        raise ValueError(f"Stimulation site {stim_site} not found in layout {stim_layout}")
+    
+    dist = np.linalg.norm(elec_pos - stim_site_pos, axis=1)
+    try:
+        idx = (dist < dist_thr[1]) & (dist >= dist_thr[0])
+    except:
+        idx = dist < dist_thr
```
rcanfield2, line 37: I think these should be <= also. Also, I think using a try/except statement is a little too general. I think it should either:
1. Use isinstance(dist_thr, tuple)/isinstance(dist_thr, float)
2. Have a way to tell the user which approach is used.

aopy/analysis/connectivity.py
leoscholl: added a link! we try not to rely on SOPs since it's a public repo

aopy/analysis/connectivity.py
leoscholl: added links

aopy/analysis/connectivity.py
leoscholl: added a better explanation. i think its typically mm but there's nothing stopping somebody from making a layout with other units

aopy/analysis/connectivity.py
leoscholl: i want to make it so that if you ask for dist_thr=[0,1] and dist_thr=[1,2] you get exclusively different results

leoscholl: i made it try / except TypeError, is that better? what specifically are you worried about when you say it's too general?

rcanfield2: In general I just don't like try/excepts because they can mask errors/bugs in the code. The user may expect that the initial approach is working, but if there is an unrelated error the function will just return the wrong answer without the user knowing. 

I prefer to condition on what is actually important. Its hard to think of an example because the error would be an unknown unknown.

rcanfield2: > i want to make it so that if you ask for dist_thr=[0,1] and dist_thr=[1,2] you get exclusively different results

Sounds good!

aopy/visualization/base.py
```diff
@@ -1934,6 +1934,81 @@ def plot_tfr(values, times, freqs, cmap='plasma', logscale=False, ax=None, **kwa
...
+            x, y = np.meshgrid(np.arange(10), np.arange(10))
+            elec_pos = np.zeros((100,2))
+            elec_pos[:,0] = x.reshape(-1)
+            elec_pos[:,1] = y.reshape(-1)
+            plot_tf_map_grid(freqs, time, tf_data, bands, elec_pos, clim=(0,1), interp_grid=None, 
+                        cmap='viridis')
+        
+        .. image:: _images/tf_map_grid.png
+    '''
+    fig, ax = plt.subplots(len(bands), len(time), figsize=(4*len(time),4*len(bands)), layout='constrained')
```
rcanfield2, line 1981: Optional argument to input ax so the user can define the plot size?

tests/test_analysis.py
```diff
@@ -1064,6 +1064,110 @@ def test_prepare_erp(self):
...
+class ConnectivityTests(unittest.TestCase):
+
+    def test_get_acq_ch_near_stimulation_site(self):
+
+        # Some sanity checks
+        fn = lambda: aopy.analysis.connectivity.get_acq_ch_near_stimulation_site(0) # no such stim site
+        self.assertRaises(ValueError, fn)
+
+        chs = aopy.analysis.connectivity.get_acq_ch_near_stimulation_site(1, dist_thr=0) # no sites within small dist
+        self.assertEqual(chs.size, 0)
```
rcanfield2, line 1076: Can you also test a distance range as input?

tests/test_analysis.py
```diff
@@ -1064,6 +1064,110 @@ def test_prepare_erp(self):
...
+
+        # Test known channels near stimulation site 14
+        chs = aopy.analysis.connectivity.get_acq_ch_near_stimulation_site(14)
+        np.testing.assert_allclose(chs, [57, 69, 70, 84])
```
rcanfield2, line 1080: Can you also test return_idx?

rcanfield2:
Looks good! Just missing a couple simple tests to verify the functions have full functionality

rcanfield2 (approved):

PR #592, opened 2024-11-19
Add code to align neuropixel insert with ecog/opto32

rcanfield2:
aopy.visualization functions still need to be updated.


rcanfield2:
Actually I'm not sure what makes the most sense to update the aopy.visualization functions.  Either:

1. Update the existing functions but will require more input args (i.e. subject)
2. Write new functions specifically for NP insert. 
3. Leave as is. (The plotting isn't difficult)

rcanfield2:
Note: I also updated `visualization.overlay_image_on_spatial_map()` to allow the user to optionally translate the image.

aopy/data/base.py
```diff
@@ -871,13 +873,6 @@ def align_recoring_drives(neuropixel_drive, drive2, subject):
...
-    # Load ch map for drives
-    np_drive_ch_pos, recording_sites , _ = load_chmap(drive_type=neuropixel_drive)
-    if drive2 == 'Opto32': 
-        drive2_ch_pos, _ , _ = load_chmap(drive_type='ECoG244') # For opto drive we still want to center the neuropixel insert relative to ECoG
-    else:
-        drive2_ch_pos, _ , _ = load_chmap(drive_type=drive2)
-    
     # If drive2 is not a neuropixel drive, load alignment angle
     config_files = files('aopy').joinpath('config')
     if drive2 not in NP_drive_list:
```
leoscholl, line 878: might want to make a note that this assumes a fixed mapping between subject and alignment which won't always be the case

aopy/data/base.py
```diff
@@ -888,16 +883,11 @@ def align_recoring_drives(neuropixel_drive, drive2, subject):
...
-    aligned_np_drive_coordinates = np.concatenate((aligned_np_drive_coordinates_x[:,None], aligned_np_drive_coordinates_y[:,None]), axis=1)
+    # Load ch map for neuropixel drive and drive 2
+    np_drive_ch_pos, recording_sites , _ = load_chmap(drive_type=neuropixel_drive, theta=-np.rad2deg(angle))    
+    drive2_ch_pos, acq_elecs , _ = load_chmap(drive_type=drive2)
 
-    # Compute offset to align with drive 2
-    drive2_offset = (np.max(drive2_ch_pos, axis=0) + np.min(drive2_ch_pos, axis=0))/2 # Find center of drive 2
-    
-    return (aligned_np_drive_coordinates + drive2_offset), recording_sites
+    return np_drive_ch_pos + neuropixel_drive_offset, drive2_ch_pos + drive2_offset, recording_sites, acq_elecs
```
leoscholl, line 890: i think this np_drive_ch_pos + offset functionality would be better handled by the load_chmap function. it could have its own 'offset' parameter. i also think that function should by default center the channel mappings on the offset (i.e. remove the rotation_offset and then add the new offset)

tests/test_data.py
```diff
@@ -1311,63 +1311,88 @@ def test_load_chmap(self):
...
 
     def test_align_recording_drives(self):
-        # Create plot for documentation with affi and biegnet
+        # Create plots for documentation with affi and biegnet
         fig, ax = plt.subplots(1,2)
         subjects = ['affi', 'beignet']
         neuropixel_drive='NP_Insert137'
         drive2 = 'ECoG244'
+        drive2_ch_pos, _ , _ = load_chmap(drive_type=drive2)
+        drive2_offset = -(np.max(drive2_ch_pos, axis=0) + np.min(drive2_ch_pos, axis=0))/2
```
leoscholl, line 1320: these offsets are pretty unintuitive. centering the maps by default would simplify things a lot

tests/test_data.py
```diff
@@ -1311,63 +1311,88 @@ def test_load_chmap(self):
...
+            visualization.base.overlay_sulci_on_spatial_map(subject, 'LM1', 'ECoG244', theta=0, offset=-drive2_offset)
+            ax.set_aspect('equal')
+            ax.set(xlim=(-8,8), ylim=(-8,8), title=f'{subject}')
             visualization.savefig(write_dir, f'{neuropixel_drive}_{drive2}_alignment_{subject}.png')
+        
+        # Test that neuropixel offset works
+        subject='affi'
+        neuropixel_drive = 'NP_Insert72'
+        drive2 = 'ECoG244'
+        aligned_np_drive_coordinates, drive2_coordinates ,  recording_sites, acq_elecs = align_recoring_drives(neuropixel_drive, drive2, subject, neuropixel_drive_offset=-drive2_offset)
```
leoscholl, line 1389: add theta=90 for affi example

tests/test_data.py
```diff
@@ -1311,63 +1311,88 @@ def test_load_chmap(self):
...
+        fig, ax = plt.subplots(1,1)
+        [ax.annotate(str(acq_elecs[ipt]), (drive2_coordinates[ipt,0], drive2_coordinates[ipt,1]), ha='center', va='center', color='k',fontsize=4) for ipt in range(len(acq_elecs))]
+        [ax.annotate(str(recording_sites[ipt]), (aligned_np_drive_coordinates[ipt,0], aligned_np_drive_coordinates[ipt,1]), ha='center', va='center', color='r', fontsize=4) for ipt in range(len(recording_sites))]
+        visualization.base.overlay_sulci_on_spatial_map(subject, 'LM1', drive2, theta=0)
+        ax.set(xlim=(-2.5,13), ylim=(-2.5,13), title=f'{subject}')
+        visualization.savefig(write_dir, f'{neuropixel_drive}_{drive2}_alignment_{subject}_npoffset.png')
```
leoscholl, line 1395: could also include these map images in the docstring

aopy/data/base.py
rcanfield2: I didn't include this because load_chmap takes in a rotation_offset parameter and I thought it would be confusing.

rcanfield2: If we should center the channel mappings on the offset we should also center on 0 by default. Either way it is altering the existing coordinates to be meaningless so we might as well remove all ambiguity.

I don't understand what you mean by remove the rotation_offset and add a new offset, I'm not applying a rotation_offset.

tests/test_data.py
rcanfield2: why? The plot it produces is correct

tests/test_data.py
rcanfield2: I don't think it makes sense to include 10 plots in the docstring. The two plots I've created are the most relevant to the data we have and if people are interested in other alignments they can generate the plots themselves

tests/test_data.py
leoscholl: to test whether rotating the whole map after alignment works

aopy/data/base.py
leoscholl: i agree centering at (0,0) makes sense by default. i think that should happen in load_chmap if possible so others can make utilities like yours without having to redo that logic. 

the rotation_offset is currently only used for rotation but the maps are left at their own coordinates (sometimes not centered on (0,0)). i suggest rotation_offset could also be subtracted permanently in load_chmap from the positions returned by `map_acq2pos`. then everything would be centered at (0,0) by default.

leoscholl:
moved to #603

PR #596, opened 2024-11-30
Neuropixel example notebook

rcanfield2:


resatomo:
Looks good. I have recently updated load_preproc_spike functions. Could you change the code so that it would use this function?

resatomo (approved):

PR #598, opened 2024-12-05
add new install requirement

katherineperks:


setup.py
```diff
@@ -28,7 +28,8 @@
     'open-ephys-python-tools',
     'aolab-bmi3d>=1.0',
     'sympy',
-    'ibllib'
+    'ibllib',
+    'seaborn>=0.13.2'
```
leoscholl, line 32: also delete the line above that just says `seaborn`

leoscholl:
can you just write a comment here for posterity that explains the requirement for >=0.13.2?

katherineperks:
Older versions of seaborn may cause this dependency error when importing aopy: 

> module `matplotlib.cm` has no attribute `register_cmap`

leoscholl (approved):

PR #603, opened 2024-12-12
add center to chmap for neuropixel drives

leoscholl:
- update load_chmap to center coordinates on each drive by default and allow positioning on a new center if desired
- simplify aligning neuropixel drives by removing the need to center 
- add ability to rotate the aligned neuropixel/drive2 together

PR #606, opened 2024-12-16
fix extents on overlay_spatial_map

leoscholl:


rcanfield2:
Can you add an offset term? So the user can shift the image around? I think you could just edit the 'x' and 'y' variable definitions like:
`   x = elec_pos[:,0] + offset[0]
    y = elec_pos[:,1] + offset[1]`

leoscholl:
yes, it's already in the other pull request #603

rcanfield2 (approved):

PR #609, opened 2024-12-17
set up skeleton jupyter notebook for a neural feature selection example

sijia66:


sijia66 (commented):
This is a code example to use lasso feature selection for selecting neural features. Some  questions to consider before merging back to the main branch. 
1. where to store the test data? 
2. SImilar functions in aopy?

leoscholl:
looks really useful, and seems that you identified some functions that could be added to aopy! i think `lasso_main`, `get_selected_unselected_electrodes` and especially `plot_electrode_selection` would be great additions; please make PRs for those separately when you have time.
i don't think you need to include the test data, that would be better suited to go along with your paper.

leoscholl (approved):
This looks useful, thanks for removing the test data. I'll make an issue that reminds us to migrate these functions into aopy!

PR #611, opened 2024-12-19
movement onset example

resatomo:
Example of reach onset and saccade onset


leoscholl:
Can you add a title at the top of the notebook (e.g. “# Movement onset example”) otherwise the table of contents will include all of the section headings

leoscholl (approved):

PR #616, opened 2024-12-19
add feature extraction example

leoscholl:


resatomo:
May be good to change name? I didn't know what features mean at first.

resatomo (approved):

leoscholl:
> May be good to change name? I didn't know what features mean at first.

i think the name that shows up is "extract neural features", does that make sense?

resatomo:
To me, I feel neural features are broad. I couldn't predict what this example notebook include when I saw the title. This example notebook is mainly for getting spectral power, right? May be good to change the name to like 'extract spectral power'?

PR #617, opened 2024-12-21
update neuropixel ipynb

rcanfield2:
Add examples to load different types of preprocessed neuropixel data.


resatomo:
Thanks for making an example. It might be good to add an explanation somewhere about `ap_samplerate` and `lfp_samplerate` so that someone would not use other sampling rate.

resatomo (approved):

rcanfield2:
I added a quick summary of the datatypes and included the sampling rate. Can you double check that it is correct? @resatomo

resatomo:
Can you add phrases like ap_samplerate and lfp_samplerate are synchronized between neuropixels and ecube, and the sampling rate is not integer, in preprocessed ap data and lfp data? But otherwise, it looks good!

PR #619, opened 2024-12-30
add options to calc_mt_tfr

leoscholl:
- default option only returns non-negative frequencies but sometimes we need them. added an option
- calc_mt_tfr only lets you input (nt, nch) data but sometimes it's useful to have (nt, nch, ntr) data. now it allows that too
- pass kwargs from calc_connectivity_map_coh to calc_connectivity_coh so we can specify the maximum frequency fk

### DIFF
--- a/aopy/analysis/base.py
+++ b/aopy/analysis/base.py
@@ -925,6 +925,11 @@ def calc_cwt_tfr(data, freqs, samplerate, fb=1.5, f0_norm=1.0, method='fft', com
         print(f"Scale ({scale[0]}, {scale[-1]})")
         print(f"Freqs ({freqs_ud[0]}, {freqs_ud[-1]})")
     
+    shape = coef.shape
+    while shape and shape[-1] == 1:
+        shape = shape[:-1] # remove trailing axes with length 1
+    coef = coef.reshape(shape)
+
     if not complex_output:
         coef = np.abs(coef)
     return freqs, time, np.flip(coef, axis=0)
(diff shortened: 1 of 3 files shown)


aopy/analysis/base.py
```diff
@@ -1093,44 +1106,63 @@ def calc_mt_tfr(ts_data, n, p, k, fs, step=None, fk=None, pad=2, ref=True, compl
         ts_data = np.array(ts_data)
     if ts_data.ndim == 1:
         ts_data = ts_data[:, np.newaxis]
+    if ts_data.ndim == 2:
+        ts_data = ts_data[:, :, np.newaxis]
     if ts_data.shape[1] == 1:
```
resatomo, line 1111: Input of [nt, nch] array is still allowed, right? Specifying this in docstrings is helpful because someone wants to still use [nt, nch] array without using trial segmented data, considering computation speed.

aopy/analysis/base.py
```diff
@@ -1214,6 +1246,11 @@ def calc_tsa_mt_tfr(data, fs, win_t, step_t, bw=None, f_max=None, pad=2, jackkni
         freqs, _win_psd, _ = calc_mt_psd(win_data, fs, bw, nfft, adaptive, jackknife, sides)
         spec[:,idx_window,...] = _win_psd
 
+    shape = spec.shape
+    while shape and shape[-1] == 1:
+        shape = shape[:-1] # remove trailing axes with length 1
+    spec = spec.reshape(shape)
+
```
resatomo, line 1253: You can use `np.squeeze()`.

resatomo (approved):

aopy/analysis/base.py
leoscholl: i think np.squeeze will remove all dimensions with length 1, so in the case with 1 channel but 2 trials it will be confusing. i only want to remove trailing dimensions.

PR #622, opened 2025-01-02
add channel mapping

leoscholl:


sijia66 (approved):
It's pretty neat that adding emg ch map needs very little code. 
It's helpful to note that the EMG channel  indices are printed on the electrode pad, where the xls file is obtained, right?

PR #626, opened 2025-01-09
add tests and funcs

rcanfield2:


aopy/analysis/behavior.py
```diff
@@ -454,4 +456,50 @@ def vector_angle(vector, in_degrees=False):
...
+                 The shape of the output will be (ntrials, ntrials).
+    '''
+
+    nt, ntrials, ndims = trajectories.shape
+    
+    traj_correlation = np.zeros((ntrials, ntrials))*np.nan
+
+    trial_variance = np.var(trajectories, axis=0) # (ntrials, ndim)
+
+    if verbose:
```
leoscholl, line 486: instead of copying the for loop block twice, you can define a variable for the iterator, something like 
```
iterator = tqdm(range(ntrials)) if verbose else range(ntrials)
for itrial in iterator:
    ...
```

aopy/analysis/behavior.py
```diff
@@ -454,4 +456,50 @@ def vector_angle(vector, in_degrees=False):
...
+                                 `ndims` is the number of dimensions for each trajectory (i.e. x, y, z).
+        center (bool): If each trial should be centered before computing correlation. (Safaie et al. 2023 sets this to true)
+        verbose (bool): If `True`, prints a progress bar during computation via the tqdm module. 
+
+    Returns:
+        ndarray: A 2D numpy array of R^2 scores between each pair of trials. 
+                 The shape of the output will be (ntrials, ntrials).
+    '''
+
+    nt, ntrials, ndims = trajectories.shape
```
leoscholl, line 480: it seems a bit strange to have the input be (nt, ntr, 3) instead of (nt, 3, ntr) since the individual trials are all (nt, 3). it would seem easier to stack trials rather than have to inserting a trial dimension in between the time and ch dimensions

tests/test_analysis.py
```diff
@@ -2069,6 +2069,25 @@ def test_vector_angle(self):
...
 
+    def test_correlate_trajectories(self):
+        ntrials = 5
+        traj = np.tile(np.arange(10)[:,None,None], (1,ntrials,2))
+        corr = aopy.analysis.behavior.correlate_trajectories(traj, verbose=False)
+        self.assertTrue(corr.shape[0]==corr.shape[0]==ntrials)
+        self.assertEqual(np.mean(corr), 1)
+
+        # Test verbososityness
+        corr = aopy.analysis.behavior.correlate_trajectories(traj, verbose=True)
```
leoscholl, line 2080: this line is copied twice for some reason

leoscholl (approved):

PR #628, opened 2025-01-10
tabulate wrapper for corners task

katherineperks:
This shouldn't break any existing uses of `get_trial_segments_and_times()`  ! Updated to be able to work with certain tasks, e.g. corners, that often have multiple of the events used as "start events" within a single trial segment. For corners, this would be the `PERIPHERAL_TARGET_ON` event.


aopy/preproc/base.py
```diff
@@ -535,6 +535,7 @@ def get_trial_segments_and_times(events, times, start_events, end_events):
         times (nt): times vector
         start_events (list): set of start events to match
         end_events (list): set of end events to match
+        repeating_start_events (bool): whether the events used as start events may occur multiple times within one segment
```
katherineperks, line 538: Explain more, maybe add an example

leoscholl, line 538:
```suggestion
        repeating_start_events (bool): whether the start events might occur multiple times within one segment. Otherwise always use the last start event within a segment. May lead to segments spanning multiple trials if used improperly. Default False.
```

aopy/preproc/base.py
```diff
@@ -535,6 +535,7 @@ def get_trial_segments_and_times(events, times, start_events, end_events):
...
 
     Returns:
         tuple: tuple containing:
```
leoscholl, line 541: would be helpful to include two examples here, one with and one without the `repeating_start_events` option, with an explanation of which events are repeating and why (multiple random targets in a trial, for instance)

leoscholl (commented):
can you add the same option to `get_trial_segments` above?

katherineperks:
Add test (check # of rows, columns in tabulate df) and add photo of the generated df to the function's docstring

aopy/data/bmi3d.py
```diff
@@ -1944,6 +1948,174 @@ def tabulate_behavior_data_out(preproc_dir, subjects, ids, dates, metadata=[],
...
+            subject = 'churro'
+            start_date = '2025-01-17'
+            end_date = '2025-01-18'
+            entries = db.lookup_mc_sessions(subject=subject, date=(date.fromisoformat(start_date), date.fromisoformat(end_date)))
+            subjects, te_ids, te_dates = db.list_entry_details(entries)
+
+            df = tabulate_behavior_data_corners(preproc_dir, subjects, te_ids, te_dates)
+            display(df.head(8))
+
+        .. image:: _images/tabulate_behavior_data_corners.png
```
katherineperks, line 2008: update image with df generated from revised function

leoscholl (approved):
this looks good. i think we should also move the handling of pause events into the main `tabulate_behavior_data` function if possible. but that can be a future PR

PR #633, opened 2025-01-25
data map correlation

leoscholl:
added a function to compute correlation across spatial maps
i also had to change a bunch of imports to avoid circularity errors


aopy/analysis/base.py
```diff
@@ -1707,6 +1708,36 @@ def calc_itpc(analytical_signals):
 '''
 Statistics
 '''
+def align_spatial_maps(data1, data2):
+    '''
+    Align two input maps by finding the location of the peak of the 2D correlation function.
+    Note, if these shifts are unexpectedly high, there is likely not high enough correlation between the maps
+    and the alignment should not be used. This function replaces input NaN values with 0 to calculate the correlation.
```
rcanfield2, line 1715: It would be nice if there was an option about how to handle the padding. (i.e. roll, zeropad, crop) However, thats probably outside the scope of this pull request. Can you just add an explanation to the documentation?

rcanfield2 (approved):

PR #635, opened 2025-02-01
Katherine tabulate tracking task

katherineperks:
Update tabulate wrapper to accommodate the newer version of tracking task that uses unique event codes to denote ramp up/ramp down times. Wrapper still works for the older version of the task that does not have these ramp-specific events.


mtringi (approved):

aopy/data/bmi3d.py
```diff
@@ -2209,32 +2211,105 @@ def tabulate_behavior_data_tracking_task(preproc_dir, subjects, ids, dates, meta
...
-
-            # If tracking successful, trajectory excludes ramp period at end of tracking
-            if new_df.loc[i,'reward']:
-                new_df.loc[i,'trajectory_end_time'] = new_df.loc[i,'tracking_end_time']-ramp_down[i]
+        tracking_start_times = event_times[np.isin(event_codes, [task_codes['CURSOR_ENTER_TARGET_RAMP_UP'],     # first occurrence is beginning of ramp up
+                                                                 task_codes['CURSOR_ENTER_TARGET']])]           # if there's no ramp up, first occurrence is beginning of trajectory
+        new_df.loc[i, 'hold_completed'] = len(tracking_start_times) > 0 # if False, HOLD_PENALTY
+        if new_df.loc[i, 'hold_completed']:
+            # Tracking begins
+            new_df.loc[i, 'tracking_start_time'] = tracking_start_times[0]
```
leoscholl, line 2257: do you know that there can't be more than one event here? might be worth adding an assertion in case things get messed up somehow

leoscholl:
would love to see the tracking task example updated with these changes!

aopy/data/bmi3d.py
katherineperks: Can you clarify? More than one event, like the list of 'tracking_start_times' is longer than 1?

katherineperks: Yeah there will likely be a bunch of events returned by this segmentation, and I only want the first one (ie first `CURSOR_ENTER_TARGET` event)

leoscholl: Ah I got it, cool

leoscholl (approved):

PR #636, opened 2025-02-05
fix decoder plot

leoscholl:


aopy/visualization/bmi3d.py
```diff
@@ -49,11 +50,12 @@ def plot_readout_map(decoder, readouts, drive_type='ECoG244', cmap='YlGnBu', ax=
     """
     if ax is None:
         fig, ax = plt.subplots()
-    channels = np.nan * np.zeros(256)
+    elec_pos, acq_ch, elecs = aodata.load_chmap(drive_type=drive_type)
+    channels = np.nan * np.zeros(np.max(acq_ch) + 1)
```
ajm2605, line 54: What are the different possible drive type arguments? I'm assuming ecog and EMG? Where would I find those

aopy/visualization/base.py
```diff
@@ -538,8 +538,8 @@ def plot_spatial_map(data_map, x, y, alpha_map=None, ax=None, cmap='bwr', nan_co
     # Calculate the proper extents
     if data_map.size > 1:
         extent = [np.min(x), np.max(x), np.min(y), np.max(y)]
-        x_spacing = (extent[1] - extent[0]) / (data_map.shape[0] - 1)
```
ajm2605, line 541: Only potential Issue I see here is if the data_map is a single column? Does that break this code? Maybe include a check if one of the dimensions are 1? maybe I don't quite understand this code though.

ajm2605 (commented):
Hey Just one question really about the data_map dimensions but I might have misunderstood the code

aopy/visualization/base.py
leoscholl: i was assuming it was 2d. I added an assertion

aopy/visualization/bmi3d.py
leoscholl: i added a pointer to load_chmap in the docstring

ajm2605 (approved):
Thanks! All makes sense to me

ajm2605 (approved):
Looks good! Did I approve the review correctly this time?

leoscholl:
> Looks good! Did I approve the review correctly this time?

Yes but I think you may not have the right permissions. Can you ask amy to give you access?

ajm2605 (approved):

ajm2605 (approved):

ajm2605 (approved):

ajm2605 (approved):

ajm2605 (approved):
did it work?

PR #643, opened 2025-02-19
Plane plotting

vaevictori:
Add code to plot planes from either transformation matrices (i.e. mappings) or equations, such as the output from linear regression.


aopy/visualization/base.py
```diff
@@ -2430,4 +2430,72 @@ def overlay_sulci_on_spatial_map(subject, chamber, drive_type, theta=0, center=(
     filename = f'{subject.lower()}_{chamber.lower()}_{drive_type.lower()}_sulci.png'
     params_file = as_file(config_dir.joinpath(filename))
     with params_file as f:
-        overlay_image_on_spatial_map(f, drive_type, theta=theta, center=center, alpha=alpha, **kwargs)
\ No newline at end of file
+        overlay_image_on_spatial_map(f, drive_type, theta=theta, center=center, alpha=alpha, **kwargs)
+
+def plot_plane(ax, plane, scale=1.0, color='grey', alpha=0.15, resolution=100):
```
leoscholl, line 2435: by convention the ax parameter is optional (default None) and if you don't supply an axis it just makes a figure with a single 3d axis.

aopy/visualization/base.py
```diff
@@ -2430,4 +2430,72 @@ def overlay_sulci_on_spatial_map(subject, chamber, drive_type, theta=0, center=(
...
+    """
+    Plots a 3D plane on a given Matplotlib axis.
+
+    This function generates a square plane centered at the origin, applies a transformation 
+    using a given mapping (either a rotation matrix or a plane equation), and renders it in 3D space.
```
leoscholl, line 2440: I think some of description is not relevant to this function
```suggestion
    Plots a 3D plane centered at the origin.
```

aopy/visualization/base.py
```diff
@@ -2430,4 +2430,72 @@ def overlay_sulci_on_spatial_map(subject, chamber, drive_type, theta=0, center=(
...
+
+    Args:
+        ax (mpl_toolkits.mplot3d.Axes3D): The Matplotlib 3D axis on which to plot the plane.
+        plane (numpy.ndarray): Specifies how the plane is transformed:
+            - If shape (3,3) or (4,4): Treated as a transformation matrix for rotation.
+            - If shape (4,): Treated as plane equation coefficients (A, B, C, D) for Ax + By + Cz + D = 0.
```
leoscholl, line 2446: can you describe in more detail what a "transformation matrix for rotation" is?
```suggestion
        plane (4-tuple or (3,3) or (4,4) matrix): Specifies how the plane is transformed:
            - If shape (3,3) or (4,4): Treated as a transformation matrix for rotating the plane z=0.
            - If shape (4,): Treated as plane equation coefficients (A, B, C, D) for Ax + By + Cz + D = 0.
```

aopy/visualization/base.py
```diff
@@ -2430,4 +2430,72 @@ def overlay_sulci_on_spatial_map(subject, chamber, drive_type, theta=0, center=(
...
+            # Example using a plane equation Ax + By + Cz + D = 0
+            plane_eq = np.array([1, 2, -1, 5])  # x + 2y - z + 5 = 0
+            plot_plane(ax, plane_eq, scale=1.0, color='red', alpha=0.5)
+
+            plt.show()
+
+        .. image:: _images/plot_plane_example.png
+    """
+    
+    xy_range = np.linspace(-10*scale, 10*scale, 100)
```
leoscholl, line 2480: i think the default scale should be 10 and here you should multiply -1*scale and 1*scale.

leoscholl, line 2480: also you're missing the `resolution` parameter here

aopy/visualization/base.py
```diff
@@ -2430,4 +2430,72 @@ def overlay_sulci_on_spatial_map(subject, chamber, drive_type, theta=0, center=(
...
+    # If plane is described as an equation:
+    elif plane.shape == (4,):
+        A,B,C,D = plane
+        z = (-A * x - B * y - D) / C
+        
+    else:
+        raise ValueError(f"Invalid mapping shape {arr.shape}. Expected (3,3) or (4,4) \
+                         for transformation matrices, or (4,) for plane equations.")
+        
+    ax.plot_surface(x, y, z, alpha=alpha, color=color)
```
leoscholl, line 2498: not necessary but maybe you could add a **kwargs parameter so that if someone wants to adjust some other parameter of plot_surface they can

aopy/visualization/base.py
vaevictori: Scale here is referring to exp_gain. Should probably change the wording to make that clear

tests/test_visualization.py
```diff
@@ -684,6 +684,22 @@ def test_gradient_trajectories(self):
...
+
+    def test_plot_plane(self):
+        from mpl_toolkits.mplot3d import Axes3D
+
+        fig = plt.figure()
+        ax = fig.add_subplot(111, projection='3d')
+
+        plane = np.eye(3)  
+        plot_plane(ax, plane, scale=1.0, color='blue', alpha=0.3)
+
```
leoscholl, line 696: this one doesn't get saved anywhere, can you add it to the docstring as well?

leoscholl: oh nevermind i get it now. disregard

aopy/visualization/base.py
leoscholl: ah ok that makes sense. maybe need to add a parameter to control the plane size too? like plane_size=10 or something?

vaevictori: Technically just deciding not to use the actual exp_gain would do this

tests/test_visualization.py
```diff
@@ -684,6 +684,22 @@ def test_gradient_trajectories(self):
...
+        # Example using a transformation matrix (identity)
+        plane = np.eye(3)  
+        plot_plane(plane, ax, gain=1.0, color='blue', alpha=0.3)
+
+        # Example using a plane equation Ax + By + Cz + D = 0
+        plane_eq = np.array([1, 2, -1, 5])  # x + 2y - z + 5 = 0
+        plot_plane(plane_eq, ax, gain=1.0, color='red', alpha=0.5)
+
+        plt.show()
```
leoscholl, line 702:
```suggestion
```

leoscholl: i think you removed this by accident

leoscholl (approved):

PR #649, opened 2025-02-27
add angle calc

leoscholl:
calculate the angle from each electrode to some origin (e.g. stim site), then bin them and calculate a statistic (mean, max, etc) across each bin


aopy/analysis/base.py
```diff
@@ -602,6 +602,61 @@ def calc_stat_over_dist_from_pos(elec_data, elec_pos, pos, statistic='mean', bin
 
     return dist, stat
 
+def calc_stat_over_angle_from_pos(elec_data, elec_pos, origin, statistic='mean', bins=20):
+    '''
+    For spatial data, calculate a statistic against the angle between each position and a given origin.
+
```
rcanfield2, line 608: Overall this function makes sense. I think the description could be more informative about the order of operations. For example, specify that it bins all data based on angle, then for data at the same angle it computes a statistic.

aopy/analysis/base.py
```diff
@@ -602,6 +602,61 @@ def calc_stat_over_dist_from_pos(elec_data, elec_pos, pos, statistic='mean', bin
...
+    Args:
+        elec_data (nelec): electrode data
+        elec_pos (nelec, 2): x, y position of each electrode
+        origin (2,): x, y position to calculate angle from
+        statistic (str): statistic to calculate ('mean', 'std', 'median', 'max', 'min'). See scipy.stats.binned_statistic. Default 'mean'.
+        bins (int or array): number of bins or bin edges for binned_statistic. Default 20.
```
rcanfield2, line 614: Specify that this is angular bins.

rcanfield2 (approved):

PR #650, opened 2025-02-28
add null stim functions

leoscholl:
- improved the tabulate_stim_data to work with multiple stim sites
- added a function to generate stimulation times
- used that function in tabulate_null_stim_data to generate timestamps across multiple recordings


aopy/data/bmi3d.py
```diff
@@ -2277,6 +2278,13 @@ def tabulate_stim_data(preproc_dir, subjects, ids, dates, metadata=['stimulation
     '''
     if df is None:
         df = pd.DataFrame()
+        
+    possible_triggers = ['qwalor_trigger', 'qwalor_ch1_trigger', 'qwalor_ch2_trigger', 
+                        'qwalor_ch3_trigger', 'qwalor_ch4_trigger']
+    possible_sensors = ['qwalor_sensor', 'qwalor_ch1_sensor', 'qwalor_ch2_sensor', 
+                       'qwalor_ch3_sensor', 'qwalor_ch4_sensor']
+    possible_stim_sites = ['stimulation_site', 'stimulation_site_ch1', 'stimulation_site_ch2',
+                           'stimulation_site_ch3', 'stimulation_site_ch4']
```
rcanfield2, line 2287: Should this be loaded from an external datafile?

aopy/data/bmi3d.py
```diff
@@ -2290,9 +2298,102 @@ def tabulate_stim_data(preproc_dir, subjects, ids, dates, metadata=['stimulation
...
+    for more information on the poisson-timed stimulation times that are generated.
+    
+    Args:
+        preproc_dir (str): base directory where the files live
+        subjects (list of str): Subject name for each recording
+        ids (list of int): Block number of Task entry object for each recording
+        dates (list of str): Date for each recording
+        metadata (list, optional): list of metadata keys that should be included in the df.
+            By default empty.
+        poisson_mu (float, optional): mean of the inter-trial times. Default 0.25.
```
rcanfield2, line 2366: Units

aopy/data/bmi3d.py
```diff
@@ -2290,9 +2298,102 @@ def tabulate_stim_data(preproc_dir, subjects, ids, dates, metadata=['stimulation
...
+        refractory_period (float, optional): minimum time between trials. Default 0.1.
+        df (DataFrame, optional): pandas DataFrame object to append. Defaults to None.
```
rcanfield2, line 2368: Add units.

aopy/data/bmi3d.py
```diff
@@ -2290,9 +2298,102 @@ def tabulate_stim_data(preproc_dir, subjects, ids, dates, metadata=['stimulation
...
+            df = pd.concat([df,pd.DataFrame(exp)], ignore_index=True)
+    
+    return df
+
+def tabulate_null_stim_data(preproc_dir, subjects, ids, dates, metadata=[], 
+                            poisson_mu=0.25, refractory_period=0.1, df=None):
+    '''
+    Concatenate null stimulation data by generating poisson-timed stimulation times
+    for the given recordings. Recordings are given as lists of subjects, task entry ids, 
+    and dates. Each list must be the same length. See :func:`~aopy.preproc.utils.generate_poisson_timestamps`
```
rcanfield2, line 2356: Is this function normally applied to data with stimulations? Do you need to exclude the the times when the laser is on then?  --- Otherwise maybe it makes more sense to title this something like "tabulate_poisson_time_points"

aopy/utils/base.py
```diff
@@ -795,4 +795,31 @@ def multiply_mat_batch(data, mat, save_path, scale = 1, max_memory_gb = 1., dtyp
...
         tmp = mat @ (scale*data[ibatch*batch_size:(ibatch+1)*batch_size, :].T)
         x[ibatch*batch_size:(ibatch+1)*batch_size, :] = (tmp.T).astype(dtype)
-        x.flush()
\ No newline at end of file
+        x.flush()
+
+def generate_poisson_timestamps(mu, max_time, min_time=0., refractory_period=0.):
+    """
+    Generate timestamps following a Poisson process with mean time between events mu,
+    with a specified minimum refractory period, and that fall within a specified time window.
```
rcanfield2, line 803: Its probably worth explicitly explaining that the number of timestamps is defined by the input parameters (mu, max_time, min_time)

aopy/utils/base.py
```diff
@@ -795,4 +795,31 @@ def multiply_mat_batch(data, mat, save_path, scale = 1, max_memory_gb = 1., dtyp
...
+    Returns:
+        np.ndarray: Array of timestamps within the specified time window.
+    """
+    # Initial estimate of required number of events (overestimate to ensure coverage)
+    estimated_events = int((max_time - min_time) / mu * 1.5)
+    
+    # Generate inter-event times
+    inter_event_times = np.random.exponential(mu, estimated_events)
+    inter_event_times = np.maximum(inter_event_times, refractory_period)
+
```
rcanfield2, line 820: In the edge case that the user inputs a refractory period >> mu, this will cease to be a poisson distribution. Its probably worth adding a warning in this case at the beginning of the function.

aopy/utils/base.py
rcanfield2: Also, that this will vary every time the function is called unless a specified seed is set.

aopy/data/bmi3d.py
leoscholl: yeah probably a good idea, i'll add something to the config folder

aopy/utils/base.py
leoscholl: yeah i'll add a note. i think even if refractory period > 0 it is no longer exactly a poisson distribution, it gets biased by expanding some of the inter_event_times to the refractory period. but this is exactly how i set it up in BMI3D so i want to keep it that way here too.

rcanfield2 (approved):

PR #655, opened 2025-03-12
classify eye behavior w remodnav

bmacag:


aopy/analysis/behavior.py
```diff
@@ -494,3 +494,310 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+    )
+    screen_half_height_deg=np.degrees(np.arctan2(screen_half_height,viewing_dist))
+    px2deg=screen_half_height_deg/screen_half_height
+    clf = remodnav.EyegazeClassifier(px2deg, samplerate, **clf_params)
+    pp = clf.preproc(data, **preproc_params)
+    events = clf(pp, classify_isp=True, sort_events=True)
+    
+    return events
+
+def get_event_start_and_end_in_eye_trajectory(eye_trajectory, event_label, clf_params, preproc_params, screen_half_height, viewing_dist, samplerate=1000):
```
leoscholl, line 596: i think it's better to not specify the default sampling rate in this function since it's critical that you enter the correct value and it will depend on the eye data

aopy/analysis/behavior.py
```diff
@@ -494,3 +494,310 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+        start_end_times (start, end): 
+            List of arrays containing start and end times of detected events for each trial in seconds.
+        start_pos (start_x, start_y):
+            List of arrays containing start positions of detected events for each trial.
+        end_pos : (end_x, end_y):
+            List of arrays containing end positions of detected events for each trial.
+    """
+
+    eye_trajectories, eye_codes = get_kinematic_segments(
+        preproc_dir, subject, te_id, date, start_events, end_events, datatype='eye'
```
leoscholl, line 661: here you can pass the sampling rate to get_kinematic_segments to force the eye_trajectories to be at the sampling rate you specify

aopy/analysis/behavior.py
```diff
@@ -494,3 +494,310 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+    for i in events:
+        if i['label'] == event_label:
+            times.append([i['start_time'],i['end_time']])
+            start_positions.append([i['start_x'], i['start_y']])
+            end_positions.append([i['end_x'], i['end_y']])
+
+    return np.array(times), np.array(start_positions), np.array(end_positions)
+
+
+def get_event_times_and_positions(preproc_dir, subject, te_id, date, start_events, end_events, event_label, clf_params, preproc_params, screen_half_height, viewing_dist, samplerate=1000):
```
leoscholl, line 632: here the sampling rate default makes sense

aopy/analysis/behavior.py
```diff
@@ -494,3 +494,310 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+    eye_trajectories, eye_codes = get_kinematic_segments(
+        preproc_dir, subject, te_id, date, start_events, end_events, datatype='eye'
+    )
+
+    start_end_times=[]
+    start_pos=[]
+    end_pos=[]
+
+    for idx, eye_trajectory in enumerate(eye_trajectories):
+        s_e, xs_ys, xe_ye = get_event_start_and_end_in_eye_trajectory(eye_trajectory, event_label, clf_params, preproc_params, screen_half_height, viewing_dist, samplerate)
```
leoscholl, line 669: do we need to check if eye_trajectories contains two eyes here?

leoscholl:
i changed my mind about where to put these functions, sorry! they should go at the bottom in aopy/precondition/eye.py
can you cut/paste them in there?

leoscholl:
i also have some suggestions for the function names:
- `get_event_start_and_end_in_eye_trajectory` could be `detect_eye_events` to differentiate it from the functions that process multiple trials of data
- `get_event_times_and_positions` and the other versions that process multiple trials could be `get_eye_event_trials`, `get_saccade_trials`, `get_fixation_trials`, etc.

or something like that..?

aopy/analysis/behavior.py
bmacag: ok that makes sense. in that case i should probably take the default out of classify_behavior_in_eye_trajectory too right?

aopy/analysis/behavior.py
bmacag: done

leoscholl:
lastly, it would be good to write a couple test functions. in tests/test_precondition.py at the bottom there is a set of EyeTests.  you could copy/paste the function called `test_detect_saccades` and replace `detect_saccades` with your single trial function (currently called `get_event_start_and_end_in_eye_trajectory`) as a starting point

aopy/analysis/behavior.py
leoscholl: yeah

aopy/analysis/behavior.py
bmacag: do you mean like add a check on the shape of the trajectory array? if so sure i can do that !

leoscholl: yes, i think you should pick the left eye here if there are two eyes. i'll make another comment but in the `classify_behavior_in_eye_trajectory` function you might also want to check that the data has already been reduced to (nt,2)

aopy/analysis/behavior.py
```diff
@@ -494,3 +494,310 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+                        - 'start_x': Start x-coordinate of the event.
+                        - 'start_y': Start y-coordinate of the event.
+                        - 'end_x': End x-coordinate of the event.
+                        - 'end_y': End y-coordinate of the event.
+                        - 'amplitude': Amplitude of the eye movement in degrees.
+                        - 'peak_velocity': Peak velocity of the eye movement in degrees/second.
+                        - 'med_velocity': Median velocity of the eye movement in degrees/second.
+                        - 'avg_velocity': Average velocity of the eye movement in degrees/second.
+    """
+    eye_data = eye_trajectory[:,:2].T
```
leoscholl, line 582: i don't know if it's a good idea to always pick the left eye here. i think it would be better to make the user input something that has the shape (nt,2) and if it is any bigger raise a ValueError with some message about the shape being wrong. you can deal with files that have two eyes in the other functions.

aopy/analysis/behavior.py
bmacag: ok! the reason i didnt pick an eye here is that it should have been taken care of already in classify_behavior_in_eye_trajectory (now called classify_event_ in_eye_trajectory). So whatever trajectories get_kinematic_segments if finding here will be formatted correctly via the get_event_start_and_end... (now called the much shorter detect_eye_events). 


So if i check the data in the first function like you suggested I wouldn't have to do anything to this function

aopy/analysis/behavior.py
bmacag: okay done!

aopy/precondition/eye.py
```diff
@@ -193,3 +195,324 @@ def detect_saccades(eye_pos, samplerate, thr=None, num_sd=1.5, intersaccade_min=
...
+        min_blink_duration=min_blink_duration,
+        dilate_nan=dilate_nan,
+        median_filter_length=median_filter_length,
+        savgol_length=savgol_length,
+        savgol_polyord=savgol_polyord,
+        max_vel=max_vel,
+        )
+    return clf_params, preproc_params
+
+def classify_event_in_eye_trajectory(eye_trajectory, clf_params, preproc_params, screen_half_height, viewing_dist, samplerate):
```
leoscholl, line 251: maybe rename to `classify_eye_events` to match the others

aopy/precondition/eye.py
```diff
@@ -193,3 +195,324 @@ def detect_saccades(eye_pos, samplerate, thr=None, num_sd=1.5, intersaccade_min=
...
+
+        Args:
+            preproc_dir (str): Base directory where file lives
+            subject (str): Subject name.
+            te_id (int): Block number of task entry object.
+            date (str): Date of the recording.
+            start_events (list): List of numeric codes representing the start of a trial.
+            end_events (list): List of numeric codes representing the end of a trial.
+            clf_params (dict): Dictionary of classifier parameters to be passed to the REMoDNaV EyegazeClassifier.
+            preproc_params (dict): Dictionary of preprocessing parameters to be passed to the classifier's preproc method.
```
leoscholl, line 476: whereever you have these arguments it would be helpful to mention the `get_default_parameters` function. if you format it as 
```
:func:`~aopy.precondition.eye.get_default_parameters`
```
it should appear as a link

leoscholl (approved):
Looks good! Thanks

PR #660, opened 2025-03-20
update task codes for auditory center out

mtringi:
add relevant codes for readyset go auditory task


mtringi:
Currently using "CUE" and "OTHER PENALTY" task codes rather than making new codes for tooslow & auditory go. Will leave pull request open a bit longer in case this needs to change based on ready set go, but I think I may end up closing the branch.

katherineperks:
Are you still using this branch? Should I review?

mtringi:
> Are you still using this branch? Should I review?

No I think the auditory cue and other penalty are enough for the task.  Let me close it

PR #666, opened 2025-04-14
allow parallel pool as input

leoscholl:
avoid some wasted time repeatedly spinning up parallel processes by letting the user manage a parallel pool themselves


rcanfield2 (approved):
Looks good. For calc_latency_st I'm just wondering if parallel should default to False since we don't have tests. (I assume mp isn't supported by unit tests?) That way, there will only be unforeseen errors if the user changes the inputs.

rcanfield2:
Looks good

PR #667, opened 2025-04-16
aba perturbation db functions

leoscholl:
added a few utilities to make finding manual control perturbation experiments easier.
- you can now call `e.get_exp_mapping()` on a task entry `e` to lookup the 3x3 mapping matrix for that experiment
- or `e.has_exp_perturbation()` to check if the mapping matrix is not the identity matrix
- added a function `get_aba_perturbation_sessions` to label recordings matching an ABA block design within a day
- added `get_aba_perturbation_days` to find any days that follow an ABA block design

i know both of you have been asking for something like this for a while, let me know if it fits your expectations or if you have any questions!


vaevictori:
Add parameter to get_exp_mapping that specifies whether the mapping should be in BMI3D (raw) or World (user) coordinates, with World as the default

vaevictori (approved):

PR #674, opened 2025-05-21
remove nans from map correlations

leoscholl:


aopy/analysis/base.py
```diff
@@ -2037,6 +2036,12 @@ def calc_spatial_map_correlation(data_maps, align_maps=False):
...
             
+    # remove NaNs so correlation doesn't output NaN
+    mask = np.any(np.isnan(flat_maps), axis=0)
+    if np.sum(mask) > 0:
+        warnings.warn(f'Removing {np.sum(mask)} NaN values in data maps')
+        flat_maps = np.array(flat_maps)[:,~mask]
+
     # Compute correlation
     flat_maps /= np.linalg.norm(flat_maps, axis=1, keepdims=True)
     NCC = np.corrcoef(flat_maps)
```
rcanfield2, line 2047: I'm sure I was the one that looked at this before but is this the same as scipy.signal.correlate2d? I worried that this isn't computing a spatial correlation. because I don't think this accounts for the 2nd dimension

leoscholl: it's not supposed to be a spatial correlation, just pairwise between same points across each map

rcanfield2: the function is called calc_spatial_map_correlation

leoscholl: haha. they do need to be spatial maps because we're optionally aligning them before doing the correlation, but yeah it's confusing. what even is a spatial correlation? would `calc_correlation_spatial_maps` be better?

rcanfield2: `calc_correlation_spatial_maps_1d`? 

Spatial correlation I think accounts for the variance in each dimension separately. (https://www.mathworks.com/help/images/ref/corr2.html#d126e63746)

leoscholl: i'm not sure my function is any different from matlab `corr2`

PR #684, opened 2025-06-30
Update wrapper proc functions

resatomo:
Updated wrapper functions for preprocessing neuropixels data


aopy/preproc/wrappers.py
```diff
@@ -379,14 +470,28 @@ def proc_lfp(data_dir, files, result_dir, result_filename, overwrite=False, max_
 
         _, lfp_metadata = aodata.filter_lfp_from_broadband(broadband_filepath, result_filepath, dtype='int16', 
                                                     max_memory_gb=max_memory_gb, **filter_kwargs)
-    elif 'ecube' in files:
+        aodata.save_hdf(result_dir, result_filename, lfp_metadata, "/lfp_metadata", append=True)
+
+    elif 'ecube' in files and 'neuropixels' not in files:
```
leoscholl, line 475: technically it is possible to have both a neuropixel probe and a ecube headstage recording at the same time. is there an easy way to change the logic to account for this?

aopy/preproc/wrappers.py
```diff
@@ -346,18 +351,104 @@ def proc_broadband(data_dir, files, result_dir, result_filename, overwrite=False
...
     Saves lfp data into the HDF datasets:
         lfp_data (nt, nch)
         lfp_metadata (dict)
     
     Args:
         data_dir (str): where the data files are located
         files (dict): dictionary of filenames indexed by system
+        result_dir (str): where to store the processed result
         result_filename (str): where to store the processed result
+        kilosort_dir (str): data folder where preprocessed data of kilosort is saved
```
leoscholl, line 451: can you explain to me what the difference between the kilosort_dir and the data_dir are?

resatomo: data_dir is `/data/raw`
kilosort_dir is `/data/preprocessed/kilosort`. Output by kilosort is saved in this folder.

resatomo:
> can you also update `proc_single` to include jobs for spikes and ap_band and in aopy.data.db make sure the `get_preprocessed_sources` is updated to include the jobs you want to run by default

I'll make a new PR for this after finishing resolving this PR.

aopy/preproc/wrappers.py
resatomo: I'll add drive number for ecube so that we can deal with simultaneous recording

aopy/preproc/wrappers.py
leoscholl: i ask about updating `proc_single` because the arguments to all the other `proc` functions is fixed and it doesn't include `kilosort_dir`

resatomo:
get_ts_data_trial and get_ts_data_segment were changed so that it would check if drives exist. Along with these changes, I added an optional argument in some tabulate functions.

aopy/preproc/wrappers.py
leoscholl: yeah you're right. let's keep it separate!

aopy/preproc/wrappers.py
```diff
@@ -507,7 +512,12 @@ def proc_lfp(data_dir, files, result_dir, result_filename, kilosort_dir=None, ov
     # Check if record_headstage is True or False
     record_headstage_key = 'record_headstage'
     metadata_group = 'exp_metadata'
-    record_headstage = aodata.load_hdf_data(result_dir, result_filename, record_headstage_key, metadata_group, cached=True)
+    tmp = result_filename.split('_')[:-1]
+    exp_filename = "_".join(tmp) + '_exp.hdf'
+    try:
+        record_headstage = aodata.load_hdf_data(result_dir, exp_filename, record_headstage_key, metadata_group, cached=True)
+    except:
+        record_headstage = True # For previous recording where there is no 'record_headstage' key
```
resatomo, line 520: It seemed like that the `record_headstage` key didn't exist when `proc_lfp` was created a long time ago. This means that `proc_lfp` is sometimes called even though the `record_headstage` doesn't exist in some old test codes. So I made an exception. I think this exception isn't weird because the `record_headstage` is always True when `proc_lfp` is called based on `preproc_job`  (`preproc_job` has the `lfp` key when `record_headstage` is `True`). So. I thought this should work. 

But in some test codes, there are some test data where the `record_headstage` exist and `record_headstage` is `False`, but `proc_lfp` is called. That's why, these try and exception code makes infinite errors in tests and currently isn't working.

Does that make sense? Do you have an alternative idea to avoid these errors?

tests/test_preproc.py
```diff
@@ -994,7 +994,7 @@ def test_parse_bmi3d_v14(self):
         self.check_required_fields(data, metadata)
         files['ecube'] = '2023-12-29_BMI3D_te13154'
         data, metadata = parse_bmi3d(data_dir, files) # with ecube data
-        self.assertEqual(len(data['sync_events']), len(data['bmi3d_events']))
+        #self.assertEqual(len(data['sync_events']), len(data['bmi3d_events']))
```
resatomo, line 997: I got this error related to this line, but I didn't know how to remove error. I commented out this line for now. Do you think this is important?
image

aopy/preproc/wrappers.py
resatomo: Thanks! This is helpful

tests/test_preproc.py
resatomo: This was because of my local environment. Now I uncommented the line.

resatomo: I don't understand why testing in local gives different results from testing in github. I'll ask you about this when I come back.

aopy/preproc/wrappers.py
```diff
@@ -346,18 +378,123 @@ def proc_broadband(data_dir, files, result_dir, result_filename, overwrite=False
...
+
+    # Check if record_headstage is True or False
+    record_headstage_key = 'record_headstage'
+    metadata_group = 'exp_metadata'
+    tmp = result_filename.split('_')[:-1]
+    exp_filename = "_".join(tmp) + '_exp.hdf'
+    record_headstage = aodata.load_hdf_data(result_dir, exp_filename, record_headstage_key, metadata_group, cached=True)
+
+    idrive = 0
+    # Preprocess neural data into lfp
```
leoscholl, line 418: I think this comment is wrong

aopy/preproc/wrappers.py
```diff
@@ -84,6 +86,32 @@ def proc_single(data_dir, files, preproc_dir, subject, te_id, date, preproc_jobs
         broadband_data, broadband_metadata = aodata.load_preproc_broadband_data(preproc_dir_base, subject, te_id, date)
         assert broadband_data.shape == (broadband_metadata['n_samples'], broadband_metadata['n_channels'])
         files['broadband'] = broadband_filename # for proc_lfp()
+    if 'spikes' in preproc_jobs:
```
leoscholl, line 89: is it possible to move `spikes` to the end? for ecube data we will want to make sure to process ap band before processing spikes

aopy/preproc/wrappers.py
```diff
@@ -346,18 +378,123 @@ def proc_broadband(data_dir, files, result_dir, result_filename, overwrite=False
...
+    # Check if record_headstage is True or False
+    record_headstage_key = 'record_headstage'
+    metadata_group = 'exp_metadata'
+    tmp = result_filename.split('_')[:-1]
+    exp_filename = "_".join(tmp) + '_exp.hdf'
+    record_headstage = aodata.load_hdf_data(result_dir, exp_filename, record_headstage_key, metadata_group, cached=True)
+
+    idrive = 0
+    # Preprocess neural data into lfp   
+    if 'neuropixels' in files and not record_headstage:
```
leoscholl, line 419: why do you care if record_headstage is true or false? if neuropixels is present then you should process it i think

leoscholl: also is there a way to check if kilosort has been run? because having 'neuropixels' in files only tells you that neuropixels data was recorded but you don't know that it's been sorted

aopy/preproc/wrappers.py
```diff
@@ -346,18 +378,123 @@ def proc_broadband(data_dir, files, result_dir, result_filename, overwrite=False
...
```
leoscholl, line 469: again i don't know why record_headstage matters

aopy/preproc/wrappers.py
```diff
@@ -372,21 +509,49 @@ def proc_lfp(data_dir, files, result_dir, result_filename, overwrite=False, max_
...
+    tmp = result_filename.split('_')[:-1]
+    exp_filename = "_".join(tmp) + '_exp.hdf'
+    try:
+        record_headstage = aodata.load_hdf_data(result_dir, exp_filename, record_headstage_key, metadata_group, cached=True)
+    except:
+        record_headstage = True # For previous recording where there is no 'record_headstage' key
+
+    idrive = 0
     # Preprocess neural data into lfp   
     if 'broadband' in files:
```
leoscholl, line 524: i think you could add `and record_headstage` here just to be safe

tests/test_data.py
```diff
@@ -643,7 +646,7 @@ def test_get_task_data(self):
 
         # Plot cycle count
         ts_data, samplerate = get_task_data(write_dir, self.subject, self.te_id, self.date, 'cycle')
-        self.assertEqual(len(ts_data), 7031)
+        self.assertEqual(len(ts_data), 7986)
```
leoscholl, line 649: can you update this figure in the docs_dir (get_cycle_data.png) that was affected by changing the test data?

PR #691, opened 2025-07-15
check length of task and clock

leoscholl:
add a preprocessing check to make sure the length of the task data and clock data are the same, and truncate otherwise

### DIFF
--- a/aopy/preproc/bmi3d.py
+++ b/aopy/preproc/bmi3d.py
@@ -420,7 +420,17 @@ def _prepare_bmi3d_v0(data, metadata):
         warnings.warn("No task data found! Cannot accurately prepare bmi3d data")
         preproc_errors.append("No task data found! Cannot accurately prepare bmi3d data")
         data['task'] = np.zeros((0,), dtype=[('time', 'f8'), ('cursor', 'f8', (3,))])
-        
+    
+    # Compare task and clock data
+    if len(data['task']) < len(data['clock']):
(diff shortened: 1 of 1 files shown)


aopy/preproc/bmi3d.py
```diff
@@ -420,7 +420,17 @@ def _prepare_bmi3d_v0(data, metadata):
...
+    
+    # Compare task and clock data
+    if len(data['task']) < len(data['clock']):
+        warnings.warn(f"Number of task cycles ({len(data['task'])}) doesn't match number of clock cycles ({len(data['clock'])}).")
+        preproc_errors.append(f"Number of task cycles ({len(data['task'])}) doesn't match number of clock cycles ({len(data['clock'])}).")
+        data['clock'] = data['clock'][:len(data['task'])]
+    elif len(data['task']) > len(data['clock']):
+        warnings.warn(f"Number of task cycles ({len(data['task'])}) is larger than number of clock cycles ({len(data['clock'])}).")
+        preproc_errors.append(f"Number of task cycles ({len(data['task'])}) is larger than number of clock cycles ({len(data['clock'])}).")
+        data['task'] = data['task'][:len(data['clock'])]
```
vaevictori, line 432: Should we consider only doing this if they are very close, so that we don't push very badly corrupted data through preprocessing? The files I'm having an issue with are consistently only one sample off, I believe.

PR #697, opened 2025-07-23
Fix trajectory plotting to accomodate flat disp

vaevictori:
Corrects assumption that first two columns should be plotted in 2D by checking for zero-columns


aopy/visualization/base.py
```diff
@@ -1163,23 +1163,60 @@ def plot_trajectories(trajectories, bounds=None, ax=None, **kwargs):
...
+
+            if np.all(np.isin(2, zero_col_idx)):
+                ax.set_xlabel('x')
+                ax.set_ylabel('y')
+            elif np.all(np.isin(1, zero_col_idx)):
+                ax.set_xlabel('x')
+                ax.set_ylabel('z')
+            else:
+                ax.set_xlabel('z')
+                ax.set_ylabel('y')
```
leoscholl, line 1207: since this code is the same for both functions can you make it its own helper function that transforms 3d data into 2d data plus an x and y label?

aopy/visualization/base.py
```diff
@@ -1300,53 +1337,103 @@ def gradient_trajectories(trajectories, n_colors=100, color_palette='viridis', b
         those bounds to manually set the first 2 axes bounds for the 3D plot.
     '''
 
+    color_list = sns.color_palette(color_palette, n_colors)
+
     if ax is None:
         ax = plt.gca()
+        
+    try: # check if 3D axes given
+        ax.set_zlabel('z')
```
leoscholl, line 1346: i think you forgot to label x and y axes for the 3d plot

leoscholl (approved):

PR #700, opened 2025-08-06
fix for random targets

leoscholl:


leoscholl:
it should work for all tasks with any number of targets!

PR #704, opened 2025-08-17
Katherine proc spikes

katherineperks:
Preproc pipeline for ecube recordings w/spikes

- filters AP band from broadband data
- saves preproc AP data & metadata
- detects spikes (threshold crossings) from AP band data
- saves preproc spike times, waveforms & metadata

These steps are wrapped in `proc_ap()` and `proc_spikes()`


aopy/data/db.py
```diff
@@ -462,6 +462,8 @@ def get_preprocessed_sources(self):
         if 'record_headstage' in params and params['record_headstage']:
             sources.append('broadband')
             sources.append('lfp')
+            sources.append('ap')
```
leoscholl, line 465: Can you make this depend on the drive type?

PR #705, opened 2025-08-19
Fxns: Initial reach & drop trials

vaevictori:
Add functions to 1) identify trials that have gaps in data over a specified time threshold; and 2) identify the index where an initial reach ends using peak-finding on movement speed. Add human center-out test data for these functions.


aopy/analysis/base.py
```diff
@@ -2561,6 +2561,180 @@ def simulate_ideal_trajectories(targets, origin=[0.0, 0.0, 0.0], resolution=1000
...
+        # 2) Compute prominences for those peaks
+        prominences, left_bases, right_bases = peak_prominences(speed, peak_idx)
+
+        # 3) Keep peaks with prominence >= (prominence_fraction * peak_height)
+        peak_heights = peak_properties["peak_heights"]
+        keep = ( prominences >= (prominence_fraction * peak_heights) )
+        peak_idx = peak_idx[keep]
+        if peak_idx.size == 0:
+            split_indices.append(len(traj))
+            continue
```
leoscholl, line 2706: can be simplified by passing the `prominence` argument to find_peaks

aopy/analysis/base.py
```diff
@@ -2561,6 +2561,180 @@ def simulate_ideal_trajectories(targets, origin=[0.0, 0.0, 0.0], resolution=1000
...
+            after = trough_idx[trough_idx > p]
+            if after.size == 0:
+                continue
+            t = after[0]
+
+            # Check the "initial peak" criterion: the trough position is >= distance_threshold
+            if np.linalg.norm(trajectory[t]) >= distance_threshold:
+                split_indices.append(int(t))
+                no_initial_peak = False
+                break
```
leoscholl, line 2728: i'm not sure i understand this part. what are you trying to do here? maybe this can be accomplished with `distance` parameter in find_peaks?

aopy/postproc/base.py
```diff
@@ -429,3 +429,116 @@ def smooth_timeseries_gaus(timeseries_data, sd, samplerate, nstd=3, conv_mode='s
...
+            if nan:
+                current_consecutive += 1
+                max_consecutive = max(max_consecutive, current_consecutive)
+            else:
+                current_consecutive = 0
+
+        if (max_consecutive / samplerate) > max_gap:
+            drop.append(True)
+        else:
+            drop.append(False)
```
leoscholl, line 542: this looks good. it maybe slow if there are a lot of trials, i suggest using the existing function from utils if you add `from .. import utils` somewhere at the top, then:
```suggestion
        max_consecutive = utils.max_repeated_nans(np.mean(trial, axis=1))
        drop.append(max_consecutive / samplerate > max_gap)
```

PR #714, opened 2025-09-30
add reindex_targets

leoscholl:
useful for long learning series with multiple experiments


aopy/utils/base.py
```diff
@@ -908,4 +911,83 @@ def scale_data_by_p_value(data, p, k=100, p0=0.08):
...
+
+def bin_vectors(vectors, start_angle=-np.pi/4, bins=4):
+    '''
+    Bin 2D vectors into angular bins.
+
+    Args:
+        vectors (ntarg): List or array of 2D vectors.
+        start_angle (float, optional): Starting angle for binning in radians. Default is -pi/4.
+        bins (int, optional): Number of angular bins. Default is 4.
+
```
rcanfield2: It would be nice to be able to control which direction around the circle that things are binned. I think you can just have a boolean arg like 'cw=True'. Then an if statement to choose between bin_edges defined in increasing or decreasing order

PR #725, opened 2025-10-30
swap channel order

leoscholl:
correct the order of channels for calc_connectivity_coh
only affects the angle output not the coherence

### DIFF
--- a/aopy/analysis/connectivity.py
+++ b/aopy/analysis/connectivity.py
@@ -142,7 +142,7 @@ def calc_connectivity_coh(data_altcond_source, data_altcond_probe, n, p, k,
         | **pair (list of tuples):** list of channel pairs
         | or **(freqs, time, coh, angle)** tuple**, if average is True
     '''
-    data_altcond = np.concatenate((data_altcond_source, data_altcond_probe), axis=1)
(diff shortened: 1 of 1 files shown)


aopy/analysis/base.py
```diff
@@ -1604,7 +1604,7 @@ def calc_mt_tfcoh(data, ch, n, p, k, fs, step, fk=None, pad=2, ref=False, imagin
             Default is False.
         imaginary (bool, optional): if True, compute imaginary coherence.
         return_angle (bool, optional): if True, also return the phase difference between 
-            the two channels. Default False.
+            the two channels. Default False. Angles correspond to phase differences from ch1 to ch0.
```
pavi-rajes: additional explanation either in this line or above in the 'ch' line would be helpful. 
Suggestions: 

> if ch = [ch1, ch0], Angles correspond to phase differences from ch1 to ch0. 
> In line 1587: Order of the channel indices impacts the sign of the angles returned in return_angle is True.

pavi-rajes (commented):
PR looks good. Added some suggestions for clarity.

pavi-rajes (approved):
Approving PR. Still highly encouraging additional explanations in doc strings. :)

PR #732, opened 2025-12-04
use circ mean

leoscholl:


ajm2605 (approved):
Maybe just add a little note in the function documentation? Otherwise yeah good catch on proper angular averaging

PR #733, opened 2026-01-12
Modified movement onset function

resatomo:
- Low-pass filter cursor position before speed is estimated because the cursor position is sometimes jittered by optitrack issues
- Update docstrings

### DIFF
--- a/aopy/analysis/behavior.py
+++ b/aopy/analysis/behavior.py
@@ -275,7 +275,7 @@ def calc_segment_duration(events, event_times, start_events, end_events, target_
 
     return segment_duration, target_codes
 
-def get_movement_onset(cursor_traj, fs, trial_start, target_onset, gocue, numsd=3.0, init_pos_trial=None, butter_order=4, low_cut=20, thr=None):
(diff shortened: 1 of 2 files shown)


aopy/analysis/behavior.py
```diff
@@ -275,40 +275,66 @@ def calc_segment_duration(events, event_times, start_events, end_events, target_
...
     movement_onset = []
     for itr in range(cursor_traj.shape[0]):
         # compute speed
-        dist = np.linalg.norm(cursor_traj[itr],axis=1)
+        filt_cursor_traj = signal.filtfilt(b1, a1, cursor_traj[itr], axis=0)
+        dist = np.linalg.norm(filt_cursor_traj-init_pos_trial[itr],axis=1)
         speed_tmp = np.diff(dist)/(1/fs)
         speed_tmp = np.insert(speed_tmp,0,speed_tmp[0]) # complement the first data point
-        speed = signal.filtfilt(b, a, speed_tmp, axis=0)
+        speed = signal.filtfilt(b2, a2, speed_tmp, axis=0)
```
mtringi: Would it be useful to also return the speed profile of each trial since it is already being calculated?

PR #737, opened 2026-01-28
katherine-update-tabulate-kinematics

katherineperks:
One small modification to `tabulate_kinematic_data()` that allows users to input a single samplerate to get all data segments or a list of samplerates that corresponds to the number of data segments. This is helpful when segmenting a session where the bmi3d frame rate really dropped off, and so using the same "idealized" samplerate for all segments will produce some segments with too many samples (relevant for spectral analyses).


katherineperks:
I need to update the docstring and maybe add an example of how to input a list of samplerates

PR #739, opened 2026-02-05
Brady engagement code

bborchelt:
Engagement code and test function.

### DIFF
--- a/aopy/analysis/behavior.py
+++ b/aopy/analysis/behavior.py
@@ -532,7 +532,7 @@ def tablet_engagement(user_traj, samplerate, time_inactive=1):
     '''
     Computes whether or not the tablet is being actively engaged by the user by looking at frames within cursor segments
     and comparing to previous frames to see if there was any change.
-    Engaged indices are labaled as 1s, while disengaged indices are labeled as 0.
+    Engaged indices are labeled as 1s, while disengaged indices are labeled as 0s.
 
     Args:
         user_traj (list of (nt,ndim) arrays): user trajectory over a trial segment; list length = n_trials, each trial is (nt, ndim)


aopy/analysis/behavior.py
```diff
@@ -527,3 +527,34 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+            # if frame is equal to prev frame: since_last += 1
+            if (frame_info == trial[frame_idx - 1]).all():
+                since_last_movement += 1
+                if since_last_movement >= frames_inactive:
+                    #add idx to bin
+                    bins.append(overall_idx)
+            else:
+                since_last_movement = 0
+            overall_idx += 1
+    return bins
```
katherineperks: It seems like the output has disengagement bins from all trials in a single list. It might be useful instead to keep the output separated by trial, so the person using the function can decide how they want to visualize the results (trial by trial vs. over a entire session).

aopy/analysis/behavior.py
```diff
@@ -527,3 +527,34 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
...
+        disengaged.
+    '''
+    
+    bins = [[] for i in range(len(user_traj))]
+
+    for trial_idx, trial in enumerate(user_traj):
+        since_last_movement = 0
+        for frame_idx, frame_info in enumerate(trial):
+            # if frame is equal to prev frame: since_last += 1
+            if (frame_info == trial[frame_idx - 1]).all():
```
katherineperks: This line breaks if the trials of `user_traj` have only 1dim (e.g. just y position). It seems like an assumption of the function that the elements of `user_traj` are (nt, ndim) where ndim > 1. Maybe you can add this to the docstring or add an assert line at the start of the function to check this is the case?

aopy/analysis/behavior.py
```diff
@@ -527,3 +527,34 @@ def correlate_trajectories(trajectories, center=True, verbose=False):
         traj_correlation[itrial,:] = np.sum(temp_corrs, axis=1) / np.sum(trial_variance[:, itrial])
     
     return traj_correlation
+
+def tablet_engagement(user_traj, frames_inactive=8):
```
katherineperks: To change this from `frames_inactive` to `time_inactive`, we'll need the samplerate of the kinematic data as an input. That way you can keep all your code the same and just add a line at the beginning with `frames_inactive = int(time_inactive*samplerate)`. 

To mirror how we're moving towards tabulating kinematic segments with each trial's actual samplerate, you might want `samplerate` to be a list of same length as `user_traj`. The calculated `frames_inactive` will also need to be a list.

katherineperks: This is a super small detail, but because you return disengaged bins, you should call this function `tablet_disengagement()`!

katherineperks (commented):
Looks great!

PR #749, opened 2026-04-10
Made some changes to summarize_entries to better suport older entries

ajm2605:
Made some changes to summarize_entries to better support older entries


leoscholl:
i think this is fixed by using the aopy.db functions to load entries instead of bmi3d db.dbfunctions
see #750

PR #756, opened 2026-08-18
Add delay periods to tabulate function of center out task

SofieDeSchrijver:
Added a helper function to get delay periods, and adjusted the center-out tabulate function to add the intended delay period per trial. Paused trials get NaN. Delay periods are discarded before August 10, 2026, because they were not/incorreclty saved before this date.


ajm2605 (approved):
Looks good

